PR #284, opened 2016-11-04
consume es6 instead of umd

tessalt:
so. 

If require.js is included on the page, then then `import ShopifyBuy` will fail if I'm pointing to the UMD build, because the umd script will be like "hey is `window.define` a thing?" and window will be like "you bet" and then it's like ok I guess i'm _not_ going to `module.exports` so then I get NOTHING. 

the es6 build is about 10kb bigger.

@tanema @michelleyschen

cc @mikkoh @minasmart

### DIFF
--- a/src/buybutton.js
+++ b/src/buybutton.js
@@ -1,4 +1,4 @@
-import ShopifyBuy from 'shopify-buy/dist/shopify-buy.umd.polyfilled';
+import ShopifyBuy from 'shopify-buy/lib/shopify-polyfilled';
 import UI from './ui';
 import productTemplates from './templates/product';
 


minasmart:
I think there's something not quite right either in how we're compiling UMD, or how browserify is consuming it.

Looking at the umd build, the conditional looks like this:

```javascript
  if (typeof define === "function" && define.amd) {
    define('shopify-buy', ['module'], factory);
```

And down an the end of the factory function, it still does `module.exports`

```javascript
  var Shopify = {
    ShopClient: ShopClient,
    Config: Config,
    version: version,
    NO_IMAGE_URI: NO_IMAGE_URI,

    buildClient: function buildClient() {
      var configAttrs = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};

      var config = new this.Config(configAttrs);

      return new this.ShopClient(config);
    }
  };

  module.exports = Shopify;
});
```

With the above in mind, I have a few questions:

- In your compiled code, what does the `import` turn into?
- And I'm also curious: Is there a way we can use the [options available](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-es2015-modules-umd#babel-plugin-transform-es2015-modules-umd) to the `umd` transform plugin to meet up with your use case better?
- This might be actually be the thing: is your browserify throwing the UMD module into a closure, so the whole thing borks, because it's double wrapped? In that case, can you set is an extern so that it doesn't get transformed?

Just so it's clear: I'm not at all against what you're doing here, but I'd like to have the UMD build working well enough so that we can document how to consume it in browserify, and you can get that 10kb back.

minasmart:
Or maybe even something simpler: If browserify is looking for common js modules that live outside a closure, and just end with a `module.exports` statement, would it be easier if we exported a rollup / commonjs module that presented a single export? 'cause that'd be easy and super quick to implement.

minasmart:
Did some investigation. Looks like [the template that babel uses](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-es2015-modules-umd/src/index.js#L17-L27) for UMD will get weird depending on what gets defined on `window`. If `window` has a keyword named `export` and you're expecting a globals, a similar thing will happen. 

I opened [an issue](https://github.com/Shopify/js-buy-sdk/issues/264) on js buy, and it should be fairly trivial to provide y'all with a commonjs build that doesn't do weird stuff.

sjcalamia:
Did this ever happen? I'm still having issues importing ShopifyBuy in webpack, and resources on the internet are slim

PR #290, opened 2016-11-08
add option to open checkout in same pane

tessalt:
allow users to pass `options.cart.popup = false` to redirect current page to checkout instead of opening a new page. 

Are there consequences to do this or reasons we didn't do this before? I don't remember. 

@tanema @michelleyschen @harisaurus @richgilbank

### DIFF
--- a/src/components/checkout.js
+++ b/src/components/checkout.js
@@ -13,6 +13,10 @@ export default class Checkout {
   }
 
   open(url) {
-    window.open(url, 'checkout', this.params);
+    if (this.config.cart.popup) {
+      window.open(url, 'checkout', this.params);
+    } else {
+      window.location = url;
+    }
   }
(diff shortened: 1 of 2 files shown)


tessalt:
So I know this _could_ impact tracking (though we don't surface those conversion stats at the moment I don't think) and yeah it's not great if people can't redirect people once they've gone through the checkout process, but some people seem to really want this and I'm inclined to just let them have it. @richgilbank @daanish any objections?

daanish:
Seems reasonable. Will this have any implications for the future if we want to do something like "in-context checkout"?

tessalt:
caveats to this option: 

- cannot redirect users back to original page after completing checkout
- cart will not be cleared on checkout
- will be unable to track successful redirect

PR #293, opened 2016-11-08
add uploader class

tessalt:
we now have versions of my original upload script across at least 4 repos. Time to squish it all into an `Uploader` class that can be customized through config options. 

I should probably add tests for this because it is not very good. 

Once this is merged we should use the same script across buy-button-storefront, js-buy-sdk, and whatever @stevebosworth was using s3 for. 

@tanema @richgilbank @minasmart @stevebosworth

### DIFF
--- a/script/util/uploader.js
+++ b/script/util/uploader.js
@@ -1,0 +1,125 @@
+"use strict";
+
+const path = require('path');
+const request = require('request');
+const fs = require('fs');
+const mime = require('mime-types');
+const exec = require('child_process').exec;
+const defaultContentType = 'application/javascript';
+
+const getContentType = function(filePath) {
+  return mime.lookup(filePath) || defaultContentType;
+}
+
+const run = function (cmd) {
+  return new Promise(function (resolve, reject) {
+    exec(cmd, function (error, stdout, stderr) {
+      if (error) {
+        reject(error);
+      }
+      if (stdout) {
+        console.log(stdout);
+      }
+      if (stderr) {
+        console.log(stderr);
+      }
+      resolve(stdout, stderr);
+    });
+  });
+}
+
+
+/**
+ * @param {Object} config
+ * @param {S3} config.s3 - aws s3 instance
+ * @param {Array} config.files - array of file paths to upload
+ * @param {String} config.dir - directory to upload all files from
+ * @param {String} config.destination - name of s3 directory to upload to
+ * @param {Number|String} config.version - version number
+ * @param {Boolean} [true] config.latest - whether to upload /latest file
+ */
+
+const Uploader = function(config) {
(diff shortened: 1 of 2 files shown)


script/util/uploader.js
```diff
@@ -0,0 +1,125 @@
...
+  return new Promise(function (resolve, reject) {
+    exec(cmd, function (error, stdout, stderr) {
+      if (error) {
+        reject(error);
+      }
+      if (stdout) {
+        console.log(stdout);
+      }
+      if (stderr) {
+        console.log(stderr);
```
minasmart: shouldn't this be `console.error` to redirect to the right fd?

script/util/uploader.js
```diff
@@ -0,0 +1,125 @@
...
+const Uploader = function(config) {
+  this.config = Object.assign({}, {
+    latest: true,
+  }, config);
+  if (config.files) {
+    this.filePaths = files;
+  }
+  if (config.dir) {
+    this.filePaths = fs.readdirSync(config.dir).map(function (file) {
+      return config.dir + '/' + file;
```
minasmart: Is there a reason to not use `path.join` and/or a template string?

script/util/uploader.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+Uploader.prototype.deployStaticFiles = function () {
+  return Promise.all(this.filePaths.map(function (filePath) {
+    const uploadPaths = [];
+    const basename = path.basename(filePath);
+    const contents = fs.readFileSync(filePath);
+
+    if (this.config.version) {
+      uploadPaths.push(this.config.destination + '/' + this.config.version + '/' + basename);
```
minasmart: `path.join` here to...?

script/util/uploader.js
```diff
@@ -0,0 +1,125 @@
...
+        uploadPaths.push(this.config.destination + '/latest/' + basename);
+      }
+    } else {
+      uploadPaths.push(this.config.destination + '/' + basename);
+    }
+
+    return Promise.all(uploadPaths.map(function (uploadPath) {
+      return this.deployStaticFile(this.config.dir + '/' + basename, uploadPath);
+    }.bind(this)));
+  }.bind(this)));
```
minasmart: What's the target node version for this uploader? Can't we use `() =>` functions?

stevebosworth:
this looks good to me, should be straight forward to update the themes deploys with this.

I'm also curious about which version of node we're on for shipit and which es6 features we can use, ran into `use strict` issues when building the themes version.  Mostly around block scoping let and const.

tessalt:
closing in favour of https://github.com/Shopify/js-uploader/pull/1

PR #297, opened 2016-11-10
[DO NOT MERGE] Money format decode

michelleyschen:
First part fix for https://github.com/Shopify/buy-button/issues/2027

Tested with the currency moneyFormat string (ie. £{{amount}})
and URI encoded format string (ie. %C2%A3%7B%7Bamount%7D%7D)

Both works!

@harisaurus @tessalt 

This depend on https://github.com/Shopify/buy-button-js/pull/299 for some edge case detecting.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -29,7 +29,7 @@ export default class Component {
     this.node = config.node;
     this.globalConfig = {
       debug: config.debug,
-      moneyFormat: config.moneyFormat || defaultMoneyFormat,
+      moneyFormat: decodeURIComponent(config.moneyFormat) || defaultMoneyFormat,
       cartNode: config.cartNode,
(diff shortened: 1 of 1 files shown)


src/styles/embeds/sass/components/_collection.scss
```diff
@@ -3,6 +3,7 @@
 }
 
 .shopify-buy__collection-products {
+  display: -webkit-box; /* Safari */
```
harisaurus: you don't need the comment :)

michelleyschen: that's actually in the wrong branch...let me clean this up!

tessalt:
👍  should be safe to merge

PR #299, opened 2016-11-10
format money better

tessalt:
make sure that empty money strings and bad money strings don't break everything. 

Also made sure money.js passed linting, and copied over the test suite from money.coffee

### DIFF
--- a/test/unit/money.js
+++ b/test/unit/money.js
@@ -1,0 +1,38 @@
+import formatMoney from '../../src/utils/money';
+
+describe('formatMoney', () => {
+  it("#format properly formats cents", () => {
+    assert.equal('0.00', formatMoney(0, '{{ amount }}'))
+    assert.equal('0.10', formatMoney(0.10, '{{ amount }}'))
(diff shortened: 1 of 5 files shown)


test/unit/money.js
```diff
@@ -0,0 +1,38 @@
...
+    assert.equal('1,010,555,523.22', formatMoney(1010555523.22, '{{ amount }}'))
+    assert.equal('1.010.555.523,22', formatMoney(1010555523.22, '{{ amount_with_comma_separator }}'))
+  });
+
+  it("#format with amount_no_decimals_with_comma_separator", () => {
+    assert.equal('1.000', formatMoney(1000.00, '{{ amount_no_decimals_with_comma_separator }}'))
+  });
+
+  it("#format with amount_no_decimals_with_space_separator", () => {
+    assert.equal('1 000', formatMoney(1000.00, '{{ amount_no_decimals_with_space_separator }}'))
```
michelleyschen: just curious, which case would something like `1 000,12` fall under. 
Or is hat something we do not support?

tessalt: I don't think so.

PR #300, opened 2016-11-11
fix tracking

tessalt:
uses correct event keys and calls `track` on view of products and collections

@richgilbank

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -375,6 +375,19 @@ export default class Product extends Component {
   }
 
   /**
+   * get info about product to be sent to tracker
+   * @return {Object}
+   */
+  get trackingInfo() {
+    return {
+      id: this.id,
+      name: this.model.selectedVariant.productTitle,
+      sku: null,
+      price: this.model.selectedVariant.price,
+    };
+  }
+
+  /**
    * get info about variant to be sent to tracker
    * @return {Object}
    */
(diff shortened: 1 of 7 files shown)


src/components/product.js
```diff
@@ -375,6 +375,19 @@ export default class Product extends Component {
   }
 
   /**
+   * get info about product to be sent to tracker
+   * @return {Object}
+   */
+  get trackingInfo() {
+    return {
+      id: this.id,
+      name: this.model.selectedVariant.productTitle,
```
richgilbank: Should it be `name` or `title`? It's `title` in product-set.

src/components/product.js
```diff
@@ -375,6 +375,19 @@ export default class Product extends Component {
...
+      sku: null,
```
richgilbank: Is SKU required?

src/components/product.js
zqureshi: I think it's called `name` here
https://github.com/Shopify/widget-framework/blob/master/app/assets/javascripts/framework/lib/tracking.js.coffee#L151

src/components/product.js
tessalt: idk in original buy button it was null if undefined

zqureshi: is it not available anymore?

reference: https://github.com/Shopify/widget-framework/blob/master/app/assets/javascripts/framework/lib/tracking.js.coffee#L150

src/utils/track.js
```diff
@@ -39,6 +39,15 @@ export default class Tracker {
     }
   }
 
+  trackComponent(type, properties) {
+    switch(type) {
+      case 'product':
+        return this.track('Viewed Product', properties);
+      case 'collection':
+        return this.track('Viewed Collection', propreties);
```
zqureshi: It's `Viewed Collections` here
https://github.com/Shopify/widget-framework/blob/master/app/assets/javascripts/framework/lib/tracking.js.coffee#L118

richgilbank:
:+1:

Issue #302, opened 2016-11-11
variant selectors too small with short text


tessalt:
![image](https://cloud.githubusercontent.com/assets/1789029/20217471/101106fa-a7ee-11e6-9721-cf2298f8342c.png)

![image](https://cloud.githubusercontent.com/assets/1789029/20217850/d0601198-a7ef-11e6-9bd0-7b18da8cc86e.png)

@harisaurus this seems to be an impact of switching variant selectors to `inline-block`

PR #304, opened 2016-11-11
fix small sizing issue for options container

harisaurus:
Fixes https://github.com/Shopify/buy-button-js/issues/302

@tanema @tessalt @michelleyschen

### DIFF
--- a/src/styles/embeds/sass/components/_product.scss
+++ b/src/styles/embeds/sass/components/_product.scss
@@ -117,6 +117,7 @@
   font-size: 14px;
 
   .shopify-buy__layout-vertical & {
+    width: 100%;
     max-width: 280px;
     display: inline-block;
   }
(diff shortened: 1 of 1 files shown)


tessalt:
does this work in the adapter case too though?

tessalt:
(lmk if you don't know how to find out)

src/styles/embeds/sass/components/_product.scss
```diff
@@ -144,7 +145,7 @@
 }
 
 .shopify-buy__btn-wrapper {
-  margin-top: $gutter;
+  margin-top: $gutter + 5;
```
michelleyschen: why do we need a + 5 again?

harisaurus:
@tessalt creating a branch to add the required fix for the adapter

src/styles/embeds/sass/components/_product.scss
michelleyschen: that would be wonderful in the future!

PR #305, opened 2016-11-14
add additional tracking props

tessalt:
adds `referrer` and `subdomain` fields to properties sent to trekkie.

@zqureshi @richgilbank @tanema @michelleyschen

### DIFF
--- a/src/utils/track.js
+++ b/src/utils/track.js
@@ -1,6 +1,7 @@
 export default class Tracker {
-  constructor(lib) {
+  constructor(lib, clientConfig) {
     this.lib = lib || null;
+    this.clientConfig = clientConfig;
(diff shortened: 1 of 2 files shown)


src/utils/track.js
```diff
@@ -49,7 +50,9 @@ export default class Tracker {
   }
 
   track(eventName, properties) {
-    properties.pageurl = document.referrer;
+    properties.pageurl = document.location.href;
+    properties.referrer = document.referrer;
+    properties.subdomain = this.clientConfig.domain;
```
zqureshi: this will be `xyz.myshopify.com`?

tessalt: yes

PR #306, opened 2016-11-15
move updateConfig methods

tessalt:
extracted `updateConfig` methods out to `Updater` classes, which components delegate to. Considered multiple approaches to this including subclassing and mixins but ultimately let's prefer composition over inheritance why not. 

### DIFF
--- a/src/updater.js
+++ b/src/updater.js
@@ -1,0 +1,18 @@
+import merge from './utils/merge';
+import Template from './template';
+
+export default class Updater {
+  constructor(component) {
+    this.component = component;
+  }
+
+  updateConfig(config) {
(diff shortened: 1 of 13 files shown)


src/updater.js
```diff
@@ -0,0 +1,18 @@
+import merge from './utils/merge';
+import Template from './template';
+
+export default class Updater {
+  constructor(component) {
+    this.component = component;
+  }
+
+  updateConfig(config) {
+    this.component.config = merge(this.component.config, config.options);
```
tanema: No more user events?

tessalt: updateConfig isn't intended for use by end users, just the admin, so I took them out.

PR #307, opened 2016-11-15
put iframe related code in view classes

tessalt:
ok lets take all the ugly iframe-related and dom-related code and dump it into some view files. 
Every Component now has a View. (they also have Models but shhhh). Views do all the gross stuff so that components can do nice things. @michelleyschen @tanema @harisaurus 

It's a big PR but it's mostly moving methods from component classes to views. 

I rewrote and updated a bunch of tests. 

ready for review!


tessalt:
really really would like to merge this please 🙏

michelleyschen:
there seems to be some conflict?

src/views/toggle.js
```diff
@@ -0,0 +1,58 @@
...
+    this.iframe.parent.setAttribute('tabindex', 0);
+    if (this.component.options.sticky) {
+      this.iframe.addClass('is-sticky');
+    }
+    if (this.isVisible) {
+      this.iframe.addClass('is-active');
+    } else {
+      this.iframe.removeClass('is-active');
+    }
+  }
```
michelleyschen: is the `this.resize()` remove from the render function for a reason?
it was added here https://github.com/Shopify/buy-button-js/pull/323/files

I actually did wanted to talk to you about if there is a better fix for this as well.

tessalt: ah yeah that's my bad in resolving the merge conflict.

PR #308, opened 2016-11-16
support reloading iframes

tessalt:
this will support those cases where the DOM is manipulated after the page is loaded, causing the iframe to reload. 

Also fun fact you can right-click on the iframes and select "reload" and now it works. 

@harisaurus @tanema @michelleyschen

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -358,7 +358,7 @@ export default class Component {
    * remove node from DOM.
    */
   destroy() {
-    this.node.parentNode.removechild(this.node);
+    this.node.parentNode.removeChild(this.node);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/component.js
```diff
@@ -205,14 +205,24 @@ export default class Component {
...
       this._userEvent('afterInit');
       return this;
     })
     .catch((err) => {
       if (err.message.indexOf('Not Found') > -1) {
         logNotFound(this);
       } else {
-        throw err;
+
+        /* eslint-disable no-console */
```
michelleyschen: If you use `//eslint-disable-next-line no-console`, then you don't have to re-enable it later!

PR #311, opened 2016-11-16
force cart toggle to not re-calculate height.

michelleyschen:
Fix https://github.com/Shopify/buy-button/issues/2050

@harisaurus @tessalt @tanema

### DIFF
--- a/src/components/toggle.js
+++ b/src/components/toggle.js
@@ -39,6 +39,10 @@ export default class CartToggle extends Component {
     return true;
   }
 
+  get outerHeight() {
+    return `${this.wrapper.clientHeight}px`;
+  }
+
   get stickyClass() {
     return this.options.sticky ? 'is-sticky' : 'is-inline';
   }


harisaurus:
@michelleyschen is it because the SVG doesn't have defined sizes perhaps? @tessalt will know better if its okay to turn off Y-recalcs

tessalt:
@harisaurus that is an excellent theory. @michelleyschen can you try defining a height on the svg and turning the recalc back on? I think it was there for a reason, possibly related to the inline version (`config.sticky = false`

michelleyschen:
SO there is currently a height of 20px set for the height of the cart icon.
Tried setting `height` directly on the SVG and it didn't work either.

It turns out that style.height in every browser return the actual height of the element, but in IE11, it return the height without the padding. Where `clientHeight` will consistently return the correct outerHeight for the toggle.

Should be pretty safe since the toggle does not resize often and even if we have more items in the cart, it would be the width that change, not the height.

tessalt:
awesome! can you test this with 
```
options: {
  toggle: {
    sticky: false
  }
}
```

airandopal:
hi there, 
Is this sticky: boolean still an option? I don't see it listed [here in the docs](http://shopify.github.io/buy-button-js/customization/#toggle-component) and I tried it and it doesn't seem to be working. I do realize this is a bit old, but I checked the current version and found [here - line 19](https://github.com/Shopify/buy-button-js/blob/d7ff71a4bccf61ccb92c69ab3e88f7b1d8e3e795/src/views/toggle.js#L19) - seems like it's checking for a config option. I'm setting it like 
```
    ui.createComponent('product', {
      id: 517777615220,
      node: document.getElementById('charger'),
      options: {
        "toggle": {
          "sticky": false,
        },
      }
    });
```
My apologies if this isn't the right place to write in about this. Feel free to point me to the appropriate channel.

PR #315, opened 2016-11-17
update defaults in bbjs to be the same with the builder

michelleyschen:
Fix https://github.com/Shopify/buy-button/issues/2063
Fix https://github.com/Shopify/buy-button/issues/2066

@harisaurus @tessalt @tanema


src/defaults/components.js
```diff
@@ -68,7 +68,7 @@ const defaults = {
       horizontal: 'shopify-buy__layout-horizontal',
     },
     text: {
-      button: 'SHOP NOW',
+      button: 'ADD TO CART',
```
richgilbank: Agree with the text change, but why is this in all caps? Should it be the default?

src/defaults/components.js
```diff
@@ -205,8 +205,8 @@ const defaults = {
     text: {
       title: 'Cart',
       empty: 'Your cart is empty.',
-      button: 'Checkout',
-      total: 'Subtotal',
+      button: 'CHECKOUT',
+      total: 'Total',
```
richgilbank: This should probably still be `Subtotal`, since discounts can be applied after this (in checkout) and this is before shipping, etc.

PR #318, opened 2016-11-18
Removing max-height in the animation make sure the full cart item app…

michelleyschen:
Fix https://github.com/Shopify/buy-button/issues/2071

@harisaurus @tessalt


tessalt:
😟  idk... have you checked your fix in some other browsers?

michelleyschen:
I think I understand the maddness that is animation now! 
TL;DR if you are using animation-fill-mode, always make sure there is a property value for start and end keyframes.

PR #319, opened 2016-11-18
don't call .contains on document

tessalt:
Apparently `.contains` does not work on `document` in IE (https://connect.microsoft.com/IE/feedback/details/780874/node-contains-is-incorrect)

This PR calls it on document.body instead.


tessalt:
@tanema @harisaurus @michelleyschen

PR #321, opened 2016-11-21
use smaller cart image

tessalt:
pull in smaller image for the cart so we're not scaling down gigantic images. 

@michelleyschen @harisaurus @tanema


tessalt:
🔔

PR #325, opened 2016-12-07
Bugfixes

tessalt:
fixes a few typos that @michelleyschen noticed, and one that was causing https://github.com/Shopify/buy-button/issues/2097#event-884527839

@michelleyschen @tanema @harisaurus

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -149,14 +149,10 @@ export default class Component {
     this._userEvent('beforeInit');
     return this.view.init().then(() => this.setupModel(data)).then((model) => {
       this.model = model;
+      this._userEvent('beforeRender');
       this.view.render();
+      this._userEvent('afterRender');
       this.view.delegateEvents();
-      if (this.iframe) {
-        this.iframe.el.onload = () => {
-          this.iframe.el.onload = null;
-          this.init();
-        };
-      }
       this._userEvent('afterInit');
       return this;
     })
--- a/src/component.js
+++ b/src/component.js
@@ -149,9 +149,7 @@ export default class Component {
     this._userEvent('beforeInit');
     return this.view.init().then(() => this.setupModel(data)).then((model) => {
       this.model = model;
-      this._userEvent('beforeRender');
       this.view.render();
-      this._userEvent('afterRender');
       this.view.delegateEvents();
       this._userEvent('afterInit');
       return this;
--- a/src/component.js
+++ b/src/component.js
@@ -103,6 +103,17 @@ export default class Component {
   }
 
   /**
+   * get styles for component and any components it contains as determined by manifest.
+   * @return {Object} key-value pairs of CSS styles.
+   */
+  get styles() {
+    return this.options.manifest.filter((component) => this.config[component].styles).reduce((hash, component) => {
+      hash[component] = this.config[component].styles;
+      return hash;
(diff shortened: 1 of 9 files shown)


src/view.js
```diff
@@ -28,6 +28,10 @@ export default class View {
       name: this.component.name,
       width: this.component.options.layout === 'vertical' ? this.component.options.width : null,
     });
+    this.iframe.el.onload = () => {
+      this.iframe.el.onload = null;
+      this.component.init();
+    };
     this.component.node.className += ` shopify-buy-frame shopify-buy-frame--${this.component.typeKey}`;
```
michelleyschen: would it matter that is may not run after `delegateEvents` anymore?

tessalt: good point I'll check

tessalt:
fixed up some further issues and did more testing

src/views/toggle.js
```diff
@@ -19,6 +19,10 @@ export default class ToggleView extends View {
     return this.component.options.sticky ? 'is-sticky' : 'is-inline';
   }
 
+  get outerHeight() {
+    return `${this.wrapper.clientHeight}px`;
+  }
+
```
michelleyschen: I think 
```
get typeKey() {		
  return 'toggle';		
}
```
is also missing?

tessalt: no that goes in `components/toggle`

tessalt:
also sneaking in a fix for an IE9 bug relating to my naive addClass implementation

michelleyschen:
mm...I am sorry but I am starting to find this really really difficult to review...

tessalt:
yeah that's because it's fixes for some confusing bugs. I'll add some line comments.

src/component.js
```diff
@@ -103,6 +103,17 @@ export default class Component {
...
+   * get styles for component and any components it contains as determined by manifest.
+   * @return {Object} key-value pairs of CSS styles.
+   */
+  get styles() {
+    return this.options.manifest.filter((component) => this.config[component].styles).reduce((hash, component) => {
+      hash[component] = this.config[component].styles;
+      return hash;
+    }, {});
+  }
+
```
tessalt: i shouldn't have moved this in the first place, it works better here.

src/utils/element-class.js
```diff
@@ -5,6 +5,10 @@ export function addClassToElement(className, element) {
   if (element.classList) {
     element.classList.add(className);
   } else {
+    const classes = element.className.split(' ');
+    if (classes.indexOf(className) > -1) {
+      return;
+    }
```
tessalt: weird event behaviour in IE9 was causing some classes to get added twice, so then not enough of them would get rmeoved. this only adds classes if they don't already exist.

harisaurus: so weird - i thought classes wouldn't double up. darn you IE

src/view.js
```diff
@@ -64,6 +64,19 @@ export default class View {
         });
       }
     });
+    if (this.iframe) {
+      this.iframe.el.onload = () => {
+        this.iframe.el.onload = null;
+        this.reloadIframe();
+      };
+    }
```
harisaurus: is there a place i can read more about why we had to do this?

PR #327, opened 2016-12-16
export css for no-iframe option

tessalt:
generates a buybutton.css file that can be pushed up to cdn and used by people not using iframes. Had to make a few changes to the code where I'd been assuming the presence of iframes. 

@michelleyschen @richgilbank

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -2,7 +2,7 @@ import morphdom from 'morphdom';
 import Template from './template';
 import Iframe from './iframe';
 import styles from './styles/embeds/all';
-import {addClassToElement} from './utils/element-class';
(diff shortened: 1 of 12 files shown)


tessalt:
also added a few lines to the docs

src/view.js
```diff
@@ -16,6 +16,7 @@ export default class View {
   }
 
   init() {
+    this.component.node.className += ` shopify-buy-frame shopify-buy-frame--${this.component.typeKey}`;
```
richgilbank: Could you use new `this.addClass` method here?

tessalt: that adds  class to the iframe, not the node (which is the div that contains it)

PR #328, opened 2016-12-16
add back missing life cycle event

michelleyschen:
add in code that was accidentally remove from https://github.com/Shopify/buy-button-js/pull/307


perrupa:
is there any chance that `component` or `component._userEvent` could be `null`?

michelleyschen:
there is always a chance, but the chance is not large :)

PR #330, opened 2016-12-20
Modal fixes

michelleyschen:
* slideOut animation was missing
* the transition is relative to `is-active` class, and therefore the `transitionend` listening should toggle `is-active` instead of `is-block`
* update doc for component that does not have its own life cycle hooks

### DIFF
--- a/docs/customization/index.md
+++ b/docs/customization/index.md
@@ -130,6 +130,8 @@ exposed on all components, several components additionally support component-spe
 The value for each event key must be a function, which will be called with one argument, which is the instance of the
 component.
 
+Note that the below hooks exist for all components except `modalProduct` and `lineItem`.
+
 ```js
 var events = {
   'beforeInit': function (component) {}, // before component is initialized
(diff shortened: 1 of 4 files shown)


docs/customization/index.md
```diff
@@ -130,6 +130,8 @@ exposed on all components, several components additionally support component-spe
 The value for each event key must be a function, which will be called with one argument, which is the instance of the
 component.
 
+Note that the below hooks exist for all components except `modalProduct` and `lineItem`.
```
richgilbank: Can you clarify why these hooks don't exist for `modalProduct` and `lineItem`?

richgilbank:
That test could be switched over to `is-block` but seems a little flakey now anyway. What's it really testing?

docs/customization/index.md
michelleyschen: both `modalProduct` and `lineItem` are not an actual component (ie. not a file in the component folder), but they are offer as component under `option` so you can still directly modify its style, content...etc.

additionally, `modalProduct` actually does have a lifecycle hook, but its the same one with `product` because internally, `modalProduct` is the same component as `product` with additional modifier.

michelleyschen:
@michelleyschen testing

PR #331, opened 2017-01-06
big update to postcss and cssnext for style transforms

harisaurus:
cssnext implements quite differently than Sass. Here's a breakdown of some of the more important findings. Will continue to expand on this, and do a FED Talks for more in depth discoveries.
1. darken/lighten sass functions are replaced with the color function that takes a color, and a `lightness()` param in a `+` or `-` percentage. for example: `color(var(--color-cart-title) lightness(+ 5%))`

2. traditional nesting in sass looks like this:
```.front {
  .end {}
  &.dev {}
}
```
which yields:
```.front .end {}
.front.dev {}
```

cssnext requires the `&` otherwise it fails, so to create the same output you’d need:
```.front {
  & .end {}
  &.dev {}
}
```
(notice the spacing between `&` and classnames

3. `&` usage is flexible in sass, such as:
```.inside {
  /* les codes */
  .outside & {}
}
```
which yields:
```.inside { /* les codes */ }
.outside .inside {}
```
cssnext `&` however are not as flexible, so you’d have to rewrite your css as follows:

```.inside {/* les codes */}
.outside {
  .inside {}
}
```

4. sass also handles calculations with less markup. i’m seeing weirdness in the syntax required for cssnext so i need to validate what is required for this. it handles additions/subtractions easier than multiplications. (edited)


src/styles/embeds/sass/components/buttons.css
```diff
@@ -30,7 +30,7 @@
 
   &:hover,
   &:focus {
-    .product__variant-img {
+    & .product__variant-img {
```
tessalt: ok so this syntax is different from sass right? in sass this would mean `.shopify-buy_btn:focus.product__variant-img` whereas this now means `.shopify-buy__btn:focus .product__variant-img` correct?

src/styles/embeds/sass/components/buttons.css
```diff
@@ -76,16 +76,16 @@
...
-  transition: transform 100ms ease;
+  transition: transform 100ms ease, color 100ms ease;
   cursor: pointer;
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
   padding-right: 9px;
 
   &:hover {
     transform: scale(1.2);
-    color: darken($color-cart-title, 5%);
+    color: color(var(--color-cart-title) lightness(+ 5%));
```
tessalt: shouldnt it be minus 5%?

src/styles/embeds/sass/components/cart.css
```diff
@@ -40,23 +42,23 @@ $cart-item-spacing: 20px;
...
   position: absolute;
   right: 0;
   box-shadow: -5px 0 5px rgba(0, 0, 0, 0.1);
 }
 
 .shopify-buy__cart__header {
-  padding: $gutter-cart;
-  padding-right: ($gutter-cart * 2);
+  padding: var(--gutter-cart);
+  padding-right: (var(--gutter-cart) * 2);
```
tessalt: this ended up working?

src/styles/embeds/sass/components/modal.css
```diff
@@ -1,21 +1,23 @@
-$modal-height: 570;
-$modal-footer-height: 90px;
+:root {
+  --modal-height: 570;
```
tessalt: why no unit?

src/styles/embeds/sass/components/buttons.css
michelleyschen: Why is & added here?

src/styles/embeds/sass/components/product.css
```diff
@@ -42,51 +44,55 @@
...
-  color: $color-black;
+  color: var(--color-black);
 
-  p, ul, ol, img {
-    margin-bottom: $gutter * (2/3);
+  & p,
+  & ul,
+  & ol,
+  & img {
+    margin-bottom: calc(var(--gutter) * (2/3));
```
michelleyschen: Does & works differently in sass vs post css and I am just confuse...?

michelleyschen: Does & works differently in sass vs post css and I am just confuse...?

src/styles/embeds/sass/components/buttons.css
```diff
@@ -76,16 +76,16 @@
   right: 9px;
   top: 8px;
   font-size: 35px;
-  color: $color-cart-title;
+  color: var(--color-cart-title);
   border: none;
   background-color: transparent;
-  transition: transform 100ms ease;
+  transition: transform 100ms ease, color 100ms ease;
```
michelleyschen: this is a bug fix

src/styles/embeds/sass/components/cart.css
```diff
@@ -1,18 +1,20 @@
-$cart-item-height: 65px;
-$cart-item-spacing: 20px;
+:root {
+  --cart-item-height: 65px;
+  --cart-item-spacing: 20px;
+}
```
michelleyschen: Talked to @harisaurus about moving all the variables to the variables file since they all complied to `:root` anyway and there is no more scoping.

He is going to do this in a separate PR after this one is out.

src/styles/embeds/sass/components/cart.css
michelleyschen: Just tested this, and this resolved to the following in css

```css
padding-right: (20px * 2);
```

src/styles/embeds/sass/components/product.css
```diff
@@ -20,20 +20,22 @@
...
+  color: var(--color-black);
+  margin-bottom: var(--gutter);
   font-weight: 700;
+}
 
-  .shopify-buy__layout-horizontal & {
-    margin-top: $gutter * (2/3);
+.shopify-buy__layout-horizontal {
+  & .shopify-buy__product__title {
+    margin-top: var(--gutter) * (2/3);
```
michelleyschen: This is resolved to the following and will probably break.

```css
margin-top: 15px * (2/3);
```

src/styles/embeds/sass/components/cart.css
michelleyschen: for simple cases that we are using variables for code readability rather than needing this value to be calculated on page render. I wonder if there will be any performance impediment?

src/styles/embeds/sass/components/product.css
```diff
@@ -157,46 +165,46 @@
...
+  & .shopify-buy__btn-wrapper {
     display: inline-block;
     vertical-align: top;
     margin: 0;
   }
 }
 
 .shopify-buy__cart-item__quantity-container {
-  margin-top: $gutter + 5;
+  margin-top: calc(var(--gutter) + 5);
```
michelleyschen: 5? 5px?

src/styles/embeds/sass/components/select.css
```diff
@@ -24,7 +24,7 @@
 
 .shopify-buy__option-select {
   & + & {
-    margin-top: $gutter / 2;
+    margin-top: var(--gutter) / 2;
```
michelleyschen: need `calc`

michelleyschen: Interesting side affect of the nesting.
this code is generating the following, which included extra empty block.

```css
.shopify-buy__option-select { } 
.shopify-buy__option-select + .shopify-buy__option-select {   margin-top: 15px / 2; }
```

src/styles/embeds/sass/components/buttons.css
harisaurus: yes, you are correct! updating

src/styles/embeds/sass/components/cart.css
harisaurus: updating

src/styles/embeds/sass/components/modal.css
harisaurus: There wasn't a unit in the original code, so I kept it consistent. Happy to add it in if you think its needed? Seems to work fine without

src/styles/embeds/sass/components/product.css
harisaurus: the value autoresolves to the measurement type, but you're right, we should be a little more specific. updating to px

harisaurus:
@tessalt @michelleyschen I've made the appropriate changes needed to ensure everything works. Another round of 👀  would be amazing. Updates for variable scoping and empty block generators will be looked at in the next PRs.

src/styles/embeds/sass/components/modal.css
tessalt: hmmmmm that seems suspicious

src/styles/embeds/sass/components/modal.css
```diff
@@ -163,16 +167,16 @@ $modal-footer-height: 90px;
 
   .shopify-buy__btn--close {
     top: -60px;
-    color: lighten($color-white, 10%);
+    color: color(var(--color-white) lightness*(+ 10%));
```
michelleyschen: Is that `*` after `lightness` meant to be there?

src/styles/embeds/sass/components/modal.css
tessalt: lets ad a unit to this.

tessalt:
just one comment

PR #335, opened 2017-01-12
update docs with popup cart option

rmnblm:
I was looking for the possibility to proceed to the checkout within the same window. After studying the code I found this undocumented option which is [checked](https://github.com/Shopify/buy-button-js/blob/2b1791de3d5f49eed55f5c1074cd980a77e5f603/src/components/checkout.js#L16) in the Checkout.js.


tessalt:
thanks!

PR #336, opened 2017-01-12
don't swallow errors

tessalt:
through a long chain of unfortunate complex things, this swallowed error was causing the "Not found" error to be incorrectly reported to bugsnag as `Cannot read property 'productTitle' of undefined`


tessalt:
THANKS GUYS

PR #337, opened 2017-01-13
add defaults and tweaks for collection product sizing and positioning

harisaurus:
add defaults and updates to collection sizings.
- ensure collection products grow when more space is available
- provide max width values for collection products
- fix up gutter spacings
- remove flexbox and use `inline-block` instead as it suffices for our requirements and prevents weird flexbox issues from popping up in older Safari versions.


tessalt:
🎩 ed 👍

harisaurus:
🔔  @tanema @michelleyschen

PR #339, opened 2017-01-18
add carousel components and functionality

harisaurus:
@tanema @tessalt @michelleyschen if I could get y'all reviewing this now that'd be great - would like to find ways to clean it up if you think there are ways to do so, and I'd like some help with improving/fixing the tests.


tanema:
It actually doesn't look that bad and I can't think of how to clean it up.

src/components/product.js
```diff
@@ -580,6 +610,40 @@ export default class Product extends Component {
     }
   }
 
+  onCarouselItemClick(evt, target) {
+    const selectedImageId = target.dataset.imageId;
```
tessalt: ie9 doesn't support dataset. use `getAttribute`

src/components/product.js
```diff
@@ -580,6 +610,40 @@ export default class Product extends Component {
...
+    const imageList = this.model.images;
+    const foundImage = imageList.find((image) => {
```
tessalt: we don't have a `find` polyfill. Use `filter()[0]`

src/components/product.js
```diff
@@ -580,6 +610,40 @@ export default class Product extends Component {
...
+    }
+    this.view.render();
+  }
+
+  onCarouselChange(toRight) {
+    const imageList = this.model.images;
+    const currentImage = imageList.find((image) => {
+      return image.id === this.currentImage.id;
+    });
+    const currentImageIndex = imageList.indexOf(currentImage);
```
tessalt: interesting approach. I _think_ it's ok. (though you'll need to replaced find with filter).

src/components/product.js
```diff
@@ -580,6 +610,40 @@ export default class Product extends Component {
...
+    if (toRight) {
```
tessalt: instead of passing through `isRight`, why not pass through `1` or `-1` ? might simplify things a bit. And maybe extract that logic into its own function so we can do early returns instead of `else`. 

```
 nextIndex(currentIndex, offset) {
  const nextIndex = currentIndex + offset;
  if (nextIndex > this.model.images.length) {
   return 0;
  }
  if (nextIndex < 0) {
    return this.model.images.length -1;
  }
  return nextIndex;
}
```

src/defaults/components.js
```diff
@@ -78,6 +83,7 @@ const defaults = {
     layout: 'horizontal',
     contents: {
       img: true,
+      // imgWithCarousel: true,
```
tessalt: ok so we need to keep `img` as a key for backwards compatibility

harisaurus:
@tessalt i've made the suggested changes - looks much cleaner, thanks! I've added additional changes to the `components` file. imgWithCarousel is disabled for everything by default as we dont want it to affect all current embeds, and will therefore allow its visibility control from within the builder

src/styles/embeds/sass/components/product.css
```diff
@@ -305,16 +305,33 @@
...
 .carousel-button {
   position: absolute;
-  top: 50%;
-  transform: translateY(-50%);
+  width: 75px;
+  top: 0;
+  height: 100%;
+  border: none;
+  font-size: 0;
+  background-image: url("http://10.91.1.169:8080/dist/arrow.svg");
```
tessalt: let's update this to what it will be once uploaded (once you're otherwise ready to merge)

harisaurus: yep yep, will do this right before merge

daanish:
This is me being really picky, but the left and right arrows stay in the hover state (dark colour) once they get clicked.

harisaurus:
@daanish yeah those are the focus states kicking in - we can explore alternates if that's feeling too weird? @andreygargul thoughts?

andreygargul:
I think the focus state is weird doesn't matter what, but we need it for a11y

daanish:
Cool, Haris explained it to me. Looks good to me!

andreygargul:
👍

richgilbank:
Is this still WIP or good to look through the code itself? 
Looking good, but a couple of things stand out to me: 

- Right now, the thumbnails are using full size images instead of thumbnail size, so load time is going to be pretty slow
- The thumbnails are background images on (mostly) empty divs that just say "true" or "false". This means they're not tabbable, and a screen reader would probably shout "false false false false true false" etc at that content
- If you open the network inspector as you cycle through the images, it looks like it keeps loading some images multiple times even though they've appeared before - are those requests necessary again or can they be cached? 
- Possible to preload the images on window load to make it feel snappier? 
- A11y - @d10k I'm suspicious of the colour contrast change for the arrows when focused/active - is it high enough contrast to be noticeable?

harisaurus:
@richgilbank:
- carousel thumbnails are now using a smaller image variant
- thumbnails have been switched to anchor tags, with their `href` set to the full sized image. this was as per feedback from the fed channel along with exploring what some of our themes do
- the images are cached after they are loaded. for some reason the network inspector lists the 2nd call made but labels the image as being retrieved from the cache
- do you mean preload all the larger images?
- andrey's handle is @andreygargul now - andrey could you look into rich's concern about colour contrast between regular + focused/active please

the code is generally good to go - we are reworking the `image` getter as it was getting out of hand. currently working on tests, but the rest should be in a reviewable state.

andreygargul:
>  I'm suspicious of the colour contrast change for the arrows when focused/active - is it high enough contrast to be noticeable?

@richgilbank depending on the photo background, arrows might not be visible at all. Saying that, I think it's fine. Here is why: 
* Arrows are not the only way to switch to the next image. You can click on thumbnail
* Arrows are not the only indicator that there is another image. We show thumbnails
* We made target area big enough that even if you just click anywhere next to the edge of the image we will switch you to the next one.
* Making arrows contrast on any background, will inevitably make them heavier. We don't want to prioritize UI (arrows) over the merchant's content (product photo)

Does it make sense?

harisaurus:
alright tests are up to date and passing - @tanema @tessalt 👀  please

harisaurus:
tested storefront also, everything looks A+

PR #344, opened 2017-01-30
turn isButton option into a getter that will check if there is a butt…

michelleyschen:
…on, the product can only be turn into a button if there is no visible button

Fix https://github.com/Shopify/buy-button/issues/2157

### DIFF
--- a/test/unit/product.js
+++ b/test/unit/product.js
@@ -300,6 +300,39 @@ describe('Product class', () => {
     });
   });
 
+  describe('get isButton', () => {
+    it('is true when isButton is turn on and there is no button', () => {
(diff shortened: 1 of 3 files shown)


test/unit/product.js
```diff
@@ -300,6 +300,39 @@ describe('Product class', () => {
     });
   });
 
+  describe('get isButton', () => {
+    it('isButton getter is true when isButton is turn on and there is no button', () => {
```
tessalt: minor nitpick: don't need to repeat getter name here (should read `it('is true...')`)

michelleyschen: okie!

PR #345, opened 2017-02-01
udpate docs with config page

tessalt:
adds a new /configuration page to the docs with a crazy huge table of ALL the possible config options. No matter what we do with the docs, we can link to this table whenever we're like "and you can find all the options here" 

added some bad css, sorry. 

@harisaurus @michelleyschen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ npm-debug.log
 /dist
 /tmp
 /lib
-index.html
+/index.html
 index.js
 legacy.html
 /src/styles/embeds/all.js
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -1,12 +1,12 @@
 <header class="docs-header">
   <ul>
     <li>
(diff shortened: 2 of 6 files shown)


michelleyschen:
This may be a stupid question, but do you guy/gal know if this file can be view as actual webpage on github or I will have to checkout the branch?

tessalt:
you'll have to check out the branch, and run `npm run docs` then visit localhost:4000/buy-button-js/configuration/

michelleyschen:
Thank you! #LearnSomethingNewEveryday

michelleyschen:
right this also mean I need to install jekyll. Do you think we should included this in the devDependency?

harisaurus:
@michelleyschen yeah its a gem, so `gem install jekyll` should suffice to install what you need and `npm run docs` will take care of running the server etc for you

michelleyschen:
@tessalt your review as PR
https://github.com/Shopify/buy-button-js/pull/347
Cheers!

PR #357, opened 2017-02-17
add chevron as img markup to prevent chrome bug

harisaurus:
A new bug in Chrome doesn't render svg background images correctly - it would show up fine in all other browsers except Chrome.

This fix moves the arrow into the actual markup as an image tag rather than a background style for the navigation buttons.


harisaurus:
got a 👍  from @tessalt IRL

PR #376, opened 2017-05-08
[WIP] Replatforming

harisaurus:
**Work In Progress**

This PR aims to replatform the `buybutton.js` library to use JS Buy SDK 1.0.0 to benefit from graphQL.


tanema:
fuck yeah

src/buybutton.js
```diff
@@ -1,16 +1,18 @@
-import ShopifyBuy from 'shopify-buy/polyfilled';
+// import ShopifyBuy from 'shopify-buy/polyfilled';
+import ShopifyBuy, {Config} from 'shopify-buy';
 import UI from './ui';
 import productTemplates from './templates/product';
 
 window.ShopifyBuy = window.ShopifyBuy || ShopifyBuy;
+window.Config = window.Config || Config;
```
tessalt: You'll definitely want to namespace this global (`Config` is pretty likely to already exist on someone's page)

src/buybutton.js
```diff
@@ -1,16 +1,18 @@
...
 
 ShopifyBuy.UI = window.ShopifyBuy.UI || {
   domains: {},
 
-  init(client, integrations = {}, styleOverrides) {
-    const domain = client.config.domain;
+  init(client, config, integrations = {}, styleOverrides) {
```
tessalt: it would be pretty great if we could keep these top-level method signatures the same. What is `config`? Does it exist on the client instance maybe?

src/buybutton.js
tanema: who is that?

src/buybutton.js
harisaurus: it used to exist on the client, but doesn't with `1.0.0` - i would've preferred to keep it the same too. @swalkinshaw @minasmart thoughts?

minasmart: The client has a subset of the previous config. This isn't a new concept, so you're totes right. keep the signature the same.

minasmart:
Closing, since there's been a workflow change.

PR #380, opened 2017-05-09
fix iframe resize based on initial image load

lyninx:
the iframe should be resizing to the height of the wrapper when its contents are loaded. this broke due to `component.model.selectedVariantImage` no longer existing on the model.

### DIFF
--- a/src/views/product.js
+++ b/src/views/product.js
@@ -1,7 +1,5 @@
 import View from '../view';
 
-const pollInterval = 200;
-
 export default class ProductView extends View {
   get className() {
     return this.component.classes.product[this.component.options.layout];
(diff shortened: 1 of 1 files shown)


src/views/product.js
```diff
@@ -16,24 +14,13 @@ export default class ProductView extends View {
   }
 
   /**
-   * check size of image until it is resolved, then set height of iframe.
+   * add event listener which triggers resize when the product image is loaded.
    */
-  resizeUntilLoaded() {
-    if (!this.iframe || !this.component.model.selectedVariantImage) {
```
harisaurus: we probably still want to include this so its not resizing when it doesn't need to?

src/views/product.js
```diff
@@ -16,24 +14,13 @@ export default class ProductView extends View {
...
-      return;
-    }
-    const img = this.wrapper.getElementsByClassName(this.component.classes.product.img)[0];
-    let intervals = 0;
-    if (img) {
-      const productResize = setInterval(() => {
```
harisaurus: Did we figure out what the polling was for? Also, does this function also address viewport resizing or is that elsewhere?

minasmart: You may have to git blame to figure out the origins of the polling.

minasmart:
Is there a test anywhere to test that this resizes on image load? If not: could you write one for an image of a known size and ensure that after everything is loaded the iframe is the right size?

src/views/product.js
tessalt: hi ok so.
I hate this. But it's the only way to get it to work consistently across all browsers. Might be worth re-testing (make sure you test FF, chrome on android, safari on iOS, with both portrait and landscape orientation images, with throttled network) to see if anything's changed, but I tried _so_ hard to avoid this. @richgilbank went through the same thing.

src/views/product.js
lyninx: ah, good point! i'll re-add it.

src/views/product.js
minasmart: What was the original issue, did `load` just not fire at the right time? or just not fire at all in some cases?

Also, thank you for providing context on this project that you're not paid to work on, you rock.

tessalt: tbh I can't remember. I _think_ the offending browser was firefox, but im not certain. And I think it was that it was firing too soon. Or it might have had something to do with cached images not firing 'load' in one of the browsers? Yeah that might have been it. So if you have an image twice on one page (which does happen) it won't fire `load` the second time.

harisaurus:
I did a bunch of testing for this and it seems to be good - I'll continue to do more device testing. I'm surprised how well this is holding up.

PR #384, opened 2017-05-17
Fix variant/option selection

lyninx:
base branch: https://github.com/Shopify/buy-button-js/pull/385

the state of the "current variant" is no longer part of the buy sdk, so we have to track it ourselves now.

this PR serves as a clean-up for much of the product class, handling variant selection functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "rollup-plugin-babel": "2.6.1",
     "rollup-plugin-commonjs": "3.3.1",
     "rollup-plugin-node-resolve": "1.7.1",
-    "shopify-buy": "^1.0.0-alpha.7",
+    "shopify-buy": "^1.0.0-alpha.8",
     "uglify-js": "2.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "rollup-plugin-babel": "2.6.1",
     "rollup-plugin-commonjs": "3.3.1",
     "rollup-plugin-node-resolve": "1.7.1",
-    "shopify-buy": "^1.0.0-alpha.8",
+    "shopify-buy": "1.0.0-alpha.8",
     "uglify-js": "2.7.0"
   }
 }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -102,7 +102,7 @@ export default class Product extends Component {
    * @return {Object} image objcet.
    */
   get currentImage() {
-    if (this.shouldUpdateImage && this.image) {
+    if (this.shouldUpdateImage) {
       this.cachedImage = this.image;
     }
 
@@ -114,11 +114,11 @@ export default class Product extends Component {
    * @return {Object} image object.
    */
   get image() {
-    if (!this.selectedVariant || (!this.selectedVariant.image && !this.options.contents.imgWithCarousel)) {
+    if (!this.selectedVariant && !this.options.contents.imgWithCarousel) {
       return null;
     }
     const imageSize = parseInt(this.options.width, 10) || 480;
-    let image;
+    let src;
     let id;
 
     const imageOptions = {
@@ -128,12 +128,12 @@ export default class Product extends Component {
 
     if (this.selectedImage) {
       id = this.selectedImage.id;
-      image = ShopifyBuy.Image.Helpers.imageForSize(this.selectedImage, imageOptions);
+      src = ShopifyBuy.Image.Helpers.imageForSize(this.selectedImage, imageOptions);
     } else {
       id = this.selectedVariant.image.id;
-      image = ShopifyBuy.Image.Helpers.imageForSize(this.selectedVariant.image, imageOptions);
+      src = ShopifyBuy.Image.Helpers.imageForSize(this.selectedVariant.image, imageOptions);
     }
-    return {id, src: image};
+    return {id, src};
   }
 
   /**
@@ -190,10 +190,7 @@ export default class Product extends Component {
       return {
         id: image.id,
         src: image.src,
-        carouselSrc: image.src, // thumbnails
-        // isSelected: image.id === this.currentImage.id,
-        // use correct image size
-        // carouselSrc: image.variants[3].src,
+        carouselSrc: image.src,
       };
     });
   }
@@ -226,13 +223,13 @@ export default class Product extends Component {
   }
 
   get variantExists() {
-    return Boolean(this.model.variants.find((variant) => {
+    return this.model.variants.some((variant) => {
       if (this.selectedVariant) {
         return variant.id === this.selectedVariant.id;
       } else {
         return false;
       }
-    }));
+    });
   }
 
   get variantInStock() {
@@ -376,9 +373,9 @@ export default class Product extends Component {
         values: option.values.map((value) => {
           return {
             name: value.value, // naming lol
-            selected: Boolean(Object.values(this.selectedOptions).find((selectedOption) => {
+            selected: Object.values(this.selectedOptions).some((selectedOption) => {
               return selectedOption === value.value;
-            })),
+            }),
           };
         }),
       };
@@ -679,7 +676,7 @@ export default class Product extends Component {
    * @return {Object} updated option object.
    */
   updateVariant(optionName, value) {
-    const updatedOption = this.model.options.filter((option) => option.name === optionName)[0];
+    const updatedOption = this.model.options.find((option) => option.name === optionName);
 
     if (updatedOption) {
       this.selectedOptions[updatedOption.name] = value;
@@ -708,7 +705,7 @@ export default class Product extends Component {
     let selectedVariant;
 
     if (this.defaultVariantId) {
-      selectedVariant = model.variants.filter((variant) => variant.id === this.defaultVariantId)[0];
+      selectedVariant = model.variants.find((variant) => variant.id === this.defaultVariantId);
     } else {
       this.defaultVariantId = model.variants[0].id;
       selectedVariant = model.variants[0];
@@ -716,9 +713,10 @@ export default class Product extends Component {
     }
 
     if (selectedVariant) {
-      selectedVariant.selectedOptions.forEach((option) => {
-        this.selectedOptions[option.name] = option.value;
-      });
+      this.selectedOptions = selectedVariant.selectedOptions.reduce((acc, option) => {
+        acc[option.name] = option.value;
+        return acc;
+      }, {});
       this.selectedVariant = selectedVariant;
     } else {
       // eslint-disable-next-line
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -1,3 +1,4 @@
+import ShopifyBuy from 'shopify-buy';
 import merge from '../utils/merge';
 import Component from '../component';
 import Template from '../template';
@@ -73,6 +74,8 @@ export default class Product extends Component {
     this.modal = null;
     this.imgStyle = '';
     this.selectedQuantity = 1;
+    this.selectedVariant = null;
+    this.selectedOptions = {};
     this.updater = new ProductUpdater(this);
     this.view = new ProductView(this);
     this.selectedImage = null;
@@ -111,10 +114,10 @@ export default class Product extends Component {
    * @return {Object} image object.
    */
   get image() {
-    if (!this.model.selectedVariant || !this.model.selectedVariant.image || !this.model.selectedVariant.image.variants.length) {
+    if (!this.selectedVariant || !this.selectedVariant.image || !this.selectedVariant.image.variants.length) {
       return null;
     }
-    const availableSizes = this.model.selectedVariant.image.variants;
+    const availableSizes = this.selectedVariant.image.variants;
 
     let imageSize = 'grande';
 
@@ -129,7 +132,7 @@ export default class Product extends Component {
       imageSize = this.options.imageSize;
     }
 
-    let sourceImage = this.model.selectedVariant.image;
+    let sourceImage = this.selectedVariant.image;
     if (this.selectedImage) {
       sourceImage = this.model.images.filter((image) => {
         return image.id === this.selectedImage.id;
@@ -148,10 +151,10 @@ export default class Product extends Component {
    * @return {String}
    */
   get formattedPrice() {
-    if (!this.model.selectedVariant) {
+    if (!this.selectedVariant) {
       return '';
     }
-    return formatMoney(this.model.selectedVariant.price, this.globalConfig.moneyFormat);
+    return formatMoney(this.selectedVariant.price, this.globalConfig.moneyFormat);
   }
 
   /**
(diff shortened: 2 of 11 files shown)


michelleyschen:
I would suggest upgrading the version of `shopify-buy` in this PR as well to address anything else that might be changing in `js-buy-sdk` since last update

lyninx:
@michelleyschen the base branch already has `"shopify-buy": "^1.0.0-alpha.7"`

michelleyschen:
ohhhhhhhhh there is a base branch

package.json
```diff
@@ -94,7 +94,7 @@
     "rollup-plugin-babel": "2.6.1",
     "rollup-plugin-commonjs": "3.3.1",
     "rollup-plugin-node-resolve": "1.7.1",
-    "shopify-buy": "^1.0.0-alpha.7",
+    "shopify-buy": "^1.0.0-alpha.8",
```
minasmart: match  exact versions.

src/components/product.js
```diff
@@ -91,16 +94,16 @@ export default class Product extends Component {
...
   }
 
   /**
    * get image for product and cache it. Return caches image if shouldUpdateImage is false.
    * @return {Object} image objcet.
    */
   get currentImage() {
-    if (this.shouldUpdateImage) {
-      this.cachedImage = this.model.images[0];
+    if (this.shouldUpdateImage && this.image) {
```
minasmart: isn't the check for `this.image` redundant here? And if not, should it not be put in the `shouldUpdateImage` predicate?

src/components/product.js
```diff
@@ -111,73 +114,62 @@ export default class Product extends Component {
...
+      maxHeight: imageSize,
+    };
 
-    let sourceImage = this.model.selectedVariant.image;
     if (this.selectedImage) {
-      sourceImage = this.model.images.filter((image) => {
-        return image.id === this.selectedImage.id;
-      })[0];
+      id = this.selectedImage.id;
+      image = ShopifyBuy.Image.Helpers.imageForSize(this.selectedImage, imageOptions);
```
minasmart: would this var be  better named as `src`? Elsewhere, `image` refers to a hash with a `src` attribute. Here, `image` refers to a literally a `src` which is returned in the statement below.

src/components/product.js
```diff
@@ -111,73 +114,62 @@ export default class Product extends Component {
...
   get viewData() {
-    const image = Object.assign({}, this.currentImage);
     return merge(this.model, this.options.viewData, {
       classes: this.classes,
       contents: this.options.contents,
       text: this.options.text,
       optionsHtml: this.optionsHtml,
       decoratedOptions: this.decoratedOptions,
-      currentImage: image,
+      currentImage: this.currentImage,
```
minasmart: Would this be better stated as `return merge(this.model, this.options.viewData, this)`? Or do all the fields need to be explicitly enumerated?

src/components/product.js
```diff
@@ -195,13 +187,14 @@ export default class Product extends Component {
...
+        id: image.id,
+        src: image.src,
+        carouselSrc: image.src, // thumbnails
         // isSelected: image.id === this.currentImage.id,
-        isSelected: true,
         // use correct image size
         // carouselSrc: image.variants[3].src,
-        carouselSrc: image.src,
-      });
+      };
```
minasmart: maybe 🔥 these comments?

src/components/product.js
```diff
@@ -233,11 +226,18 @@ export default class Product extends Component {
   }
 
   get variantExists() {
-    return Boolean(this.model.selectedVariant);
+    return Boolean(this.model.variants.find((variant) => {
```
minasmart: I think you could skip the explicit  `Boolean` and just swap `.find` for `.some`

src/components/product.js
```diff
@@ -385,16 +370,15 @@ export default class Product extends Component {
...
       return {
         name: option.name,
         values: option.values.map((value) => {
           return {
-            name: value,
-            selected: value === option.selected,
-            disabled: false,
-            // disabled: !this.optionValueCanBeSelected(selections, option.name, value),
+            name: value.value, // naming lol
+            selected: Boolean(Object.values(this.selectedOptions).find((selectedOption) => {
```
minasmart: `.some` No need for `Boolean`

src/components/product.js
```diff
@@ -406,14 +390,14 @@ export default class Product extends Component {
...
    */
   get trackingInfo() {
-    if (!this.model.selectedVariant) {
+    if (!this.selectedVariant) {
       return {};
     }
     return {
       id: this.id,
-      name: this.model.selectedVariant.productTitle,
+      name: this.selectedVariant.productTitle,
```
minasmart: why call this a `name` if it's called a `title`?

src/components/product.js
```diff
@@ -695,10 +680,21 @@ export default class Product extends Component {
    */
   updateVariant(optionName, value) {
     const updatedOption = this.model.options.filter((option) => option.name === optionName)[0];
```
minasmart: can this not be `find`?

src/components/product.js
```diff
@@ -709,17 +705,22 @@ export default class Product extends Component {
    * @param {Object} model - model to be modified.
    */
   setDefaultVariant(model) {
-    if (!this.defaultVariantId) {
-      return model;
+    let selectedVariant;
+
+    if (this.defaultVariantId) {
+      selectedVariant = model.variants.filter((variant) => variant.id === this.defaultVariantId)[0];
```
minasmart: `.find`?

src/components/product.js
```diff
@@ -709,17 +705,22 @@ export default class Product extends Component {
...
+      selectedVariant = model.variants[0];
+      this.selectedImage = model.images[0];
     }
 
-    const selectedVariant = model.variants.filter((variant) => variant.id === this.defaultVariantId)[0];
     if (selectedVariant) {
-      model.options.forEach((option) => {
-        option.selected = selectedVariant.optionValues.filter((optionValue) => optionValue.name === option.name)[0].value;
+      selectedVariant.selectedOptions.forEach((option) => {
+        this.selectedOptions[option.name] = option.value;
```
minasmart: could this be written as a reduce?

```javascript
this.selectedOptions = selectedVariant.selectedOptions.reduce((acc,  option) => {
  acc[option.name] = option.value;
}, {});
```

src/views/product.js
```diff
@@ -19,7 +19,7 @@ export default class ProductView extends View {
    * check size of image until it is resolved, then set height of iframe.
    */
   resizeUntilLoaded() {
-    if (!this.iframe || !this.component.model.selectedVariantImage) {
+    if (!this.iframe || !this.component.currentImage) {
```
minasmart: `!(this.iframe && this.component.currentImage)` may read more clear? To me, this version looks like `neither A nor B`, as opposed to `not A or not B` which is really confusing.

test/unit/buybutton.js
```diff
@@ -1,17 +1,16 @@
...
 
-describe.skip('ShopifyBuy.UI', () => {
-  const configAttrs = {
+describe('ShopifyBuy.UI', () => {
+  const config = new Config({
     domain: 'embeds.myshopify.com',
-    apiKey: 123,
-    appId: 6
-  };
+    storefrontAccessToken: 'meow',
```
minasmart: maybe use a value that's more clearly an access token. Like `access-token-12345` or something?

test/unit/product.js
```diff
@@ -171,8 +171,9 @@ describe('Product class', () => {
         });
       });
       describe('if variant is not in stock', () => {
+
```
minasmart: 🔥

test/unit/product.js
```diff
@@ -337,7 +333,7 @@ describe('Product class', () => {
     it('it updates selected variant', () => {
       return product.init(testProductCopy).then(() => {
         let updated = product.updateVariant('Size', 'large');
-        assert.equal(updated.selected, 'large');
+        assert.equal(product.selectedOptions['Size'], 'large');
```
minasmart: this is better written as `product.selectedOptions.Size`

test/unit/product.js
```diff
@@ -402,7 +398,7 @@ describe('Product class', () => {
       assert.calledWith(cartUpdateConfigSpy, newConfig);
     });
 
-    it('calls updateConfig on modal if modal exists', () => {
+    it.skip('calls updateConfig on modal if modal exists', () => {
```
minasmart: Not asking for a change here, just want to sure the `.skip` is intentional.

test/unit/product.js
```diff
@@ -450,8 +446,8 @@ describe('Product class', () => {
     it('sets selectedVariant to product.defalutVariantId', () => {
       product.defaultVariantId = 12347;
       const model = product.setDefaultVariant(testProduct);
-      assert.equal(model.options[0].selected, 'shark');
-      assert.equal(model.options[1].selected, 'large');
+      assert.equal(product.selectedOptions['Print'], 'shark');
```
minasmart: Use dot notation where you can.

test/unit/product.js
```diff
@@ -695,30 +669,30 @@ describe('Product class', () => {
...
+  describe('onCarouselChange', () => {
     beforeEach(() => {
       return product.init(testProductCopy).then(() => {
         return Promise.resolve();
       });
     });
     it('sets selected image based on various offsets', () => {
       product.onCarouselChange(-1);
-      assert.equal(product.image.src, 'https://cdn.shopify.com/image-four_large.jpg');
+      assert.equal(product.image.src, 'https://cdn.shopify.com/s/image-four_280x280.jpg');
```
minasmart: maybe stick the base URI in a var?

src/components/product.js
lyninx: that would simplify the code, but it just means including a bunch of other fluff in viewData which it doesn't need.

src/components/product.js
lyninx: neat! i didn't know `.some` was a thing! thanks 👍

src/components/product.js
lyninx: ah yes! good catch. i like the reduce better.

test/unit/product.js
lyninx: it is! this test passes, but it blocks because of the modal initialization step which will be fixed later on.

src/components/product.js
```diff
@@ -195,13 +188,11 @@ export default class Product extends Component {
 
   get carouselImages() {
     return this.model.images.map((image) => {
-      return Object.assign({}, image, {
-        // isSelected: image.id === this.currentImage.id,
-        isSelected: true,
-        // use correct image size
-        // carouselSrc: image.variants[3].src,
```
harisaurus: Assuming these things are being addressed in the carousel work?

src/components/product.js
```diff
@@ -233,11 +224,18 @@ export default class Product extends Component {
...
+        return variant.id === this.selectedVariant.id;
+      } else {
+        return false;
+      }
+    });
   }
 
   get variantInStock() {
-    return this.variantExists && this.model.selectedVariant.available;
+    // variant availability is no longer provided as part of the api?
```
harisaurus: 🔥

src/components/product.js
```diff
@@ -385,16 +368,15 @@ export default class Product extends Component {
...
       return {
         name: option.name,
         values: option.values.map((value) => {
           return {
-            name: value,
-            selected: value === option.selected,
-            disabled: false,
-            // disabled: !this.optionValueCanBeSelected(selections, option.name, value),
+            name: value.value, // naming lol
+            selected: Object.values(this.selectedOptions).some((selectedOption) => {
```
harisaurus: what's the difference between `.some` and `.filter` and why did we opt for `.some`?

src/components/product.js
```diff
@@ -619,20 +601,21 @@ export default class Product extends Component {
     evt.preventDefault();
     const selectedImageId = target.getAttribute('data-image-id');
     const imageList = this.model.images;
-    const foundImage = imageList.filter((image) => {
-      return image.id === parseInt(selectedImageId, 10);
-    })[0];
+    const foundImage = imageList.find((image) => {
```
harisaurus: noice

src/components/product.js
lyninx: `.some` returns a boolean, so it saves us having to wrap with `Boolean()`

src/components/product.js
lyninx: A lot of this is no longer required because we now fetch resized images with `ShopifyBuy.Image.Helpers.imageForSize`, we we just have to keep track of a single image GraphModel for what's currently selected

harisaurus: Awesome - we'll have to remember to clean up cruft once we've got everything sorted.

src/components/product.js
harisaurus: Nice! hadn't seen `.some` used before. Very cool

src/components/product.js
```diff
@@ -111,76 +114,66 @@ export default class Product extends Component {
    * @return {Object} image object.
    */
   get image() {
-    if (!this.model.selectedVariant || !this.model.selectedVariant.image || !this.model.selectedVariant.image.variants.length) {
+    if (!this.selectedVariant && !this.options.contents.imgWithCarousel) {
```
minasmart: I think this may be better written as a neither `(!(A && B))` thoughts?

src/components/product.js
```diff
@@ -195,13 +188,11 @@ export default class Product extends Component {
...
-        // isSelected: image.id === this.currentImage.id,
-        isSelected: true,
-        // use correct image size
-        // carouselSrc: image.variants[3].src,
+      return {
+        id: image.id,
+        src: image.src,
         carouselSrc: image.src,
-      });
+      };
```
minasmart: are `src` and `carouselSrc` both needed?

src/components/product.js
```diff
@@ -385,16 +368,15 @@ export default class Product extends Component {
...
     return this.model.options.map((option) => {
       return {
         name: option.name,
         values: option.values.map((value) => {
           return {
-            name: value,
-            selected: value === option.selected,
-            disabled: false,
-            // disabled: !this.optionValueCanBeSelected(selections, option.name, value),
+            name: value.value, // naming lol
```
minasmart: this comment can probably go.

src/components/product.js
```diff
@@ -406,14 +388,14 @@ export default class Product extends Component {
    * @return {Object}
    */
   get trackingInfo() {
-    if (!this.model.selectedVariant) {
+    if (!this.selectedVariant) {
```
minasmart: may be a little more straight forward to avoid the negated condition. Just `if (this.selectedVariant) { return { stuff } } else { return  {} }`

src/components/product.js
```diff
@@ -619,20 +601,21 @@ export default class Product extends Component {
...
       this.cachedImage = foundImage;
     }
+
     this.view.render();
   }
 
   nextIndex(currentIndex, offset) {
     const nextIndex = currentIndex + offset;
-    if (nextIndex > this.model.images.length) {
+    if (nextIndex > this.model.images.length - 1) {
```
minasmart: couldn't this be more easily written as `if (nextIndex === this.model.images.length)` and skip the math? Or if you're paranoid, use `>=`?

src/components/product.js
lyninx: hmmm. maybe. i think i was actually planning on using resized images for `carouselSrc`, but forgot to get back to that. good catch!

src/components/product.js
lyninx: sure! that makes it easier to understand. i'll change it

src/components/product.js
lyninx: ah, right! that makes much more sense 👍

PR #389, opened 2017-06-08
fix cart + checkout

lyninx:
fixing how products are added, updated, and removed form the cart, as well as how a checkout is generated.

since there is a perceptible delay when updating items in the cart, my next PR will contain performance improvements (caching + client side calculations), for modifying the cart.

### DIFF
--- a/src/components/cart.js
+++ b/src/components/cart.js
@@ -1,3 +1,4 @@
+import ShopifyBuy from 'shopify-buy';
 import merge from '../utils/merge';
 import Component from '../component';
 import CartToggle from './toggle';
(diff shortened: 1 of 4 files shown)


src/components/cart.js
```diff
@@ -77,8 +78,8 @@ export default class Cart extends Component {
       const data = merge(lineItem, this.options.viewData);
       data.classes = this.classes;
       data.lineItemImage = this.imageForLineItem(data);
-      data.variantTitle = data.variant_title === 'Default Title' ? '' : data.variant_title;
-      data.formattedPrice = formatMoney(data.line_price, this.globalConfig.moneyFormat);
+      data.variantTitle = data.variant.title === 'Default Title' ? '' : data.variant.title;
+      data.formattedPrice = formatMoney(data.variant.price * data.quantity, this.globalConfig.moneyFormat);
```
minasmart: this should rely on the shop resource. Is that's what's driving `globalConfig` now?

test/unit/cart.js
```diff
@@ -67,7 +70,8 @@ describe('Cart class', () => {
             title: 'test',
             variant_title: 'test2',
             line_price: 20,
-            quantity: 1
+            quantity: 1,
+            variant: { image: { src: 'cdn.shopify.com/image.jpg' } }
```
minasmart: use braces consistently. Nowhere else is are you doing `{ stuff` it's always `{stuff`. This should probably be in the linter.

lyninx: ah. i always default to adding spaces, and then fix it when the linter yells at me. i'm surprised it didn't catch this.

src/components/cart.js
lyninx: gotchya. i'll look into updating this.

lyninx:
i'll be sure to take care of fixing linting of tests in a future PR

PR #390, opened 2017-06-13
performance fixes

lyninx:
fixing performance, to let the graphQL storefront api version of buy button be at least as responsive as the previous version.

will include any random clean up work in this PR as well.


src/components/cart.js
```diff
@@ -27,6 +27,7 @@ export default class Cart extends Component {
     this.childTemplate = new Template(this.config.lineItem.templates, this.config.lineItem.contents, this.config.lineItem.order);
     this.node = config.node || document.body.appendChild(document.createElement('div'));
     this.isVisible = this.options.startOpen;
+    this.cache = null;
```
minasmart: more descriptive name. Without reading more coed, I don't know what's being cached here. If it's line items, maybe `lineItemCache`?

src/components/cart.js
```diff
@@ -74,12 +75,15 @@ export default class Cart extends Component {
...
-    return this.model.lineItems.reduce((acc, lineItem) => {
-      const data = merge(lineItem, this.options.viewData);
+    return this.cache.reduce((acc, lineItem) => {
+      const data = Object.assign({}, lineItem, this.options.viewData);
       data.classes = this.classes;
       data.lineItemImage = this.imageForLineItem(data);
       data.variantTitle = data.variant.title === 'Default Title' ? '' : data.variant.title;
-      data.formattedPrice = formatMoney(data.variant.price * data.quantity, this.globalConfig.moneyFormat);
+      if (this.config.moneyFormat) {
+        this.moneyFormat = this.globalConfig.moneyFormat;
```
minasmart: why `config` in the if and `globalConfig' in the assignment?

src/components/cart.js
```diff
@@ -220,21 +237,57 @@ export default class Cart extends Component {
...
+        const updatedItem = this.cache.find((cacheItem) => {
+          return cacheItem.id === item.id;
+        });
+        return Object.assign({}, updatedItem, item);
+      });
+    } else {
+      this.cache = lineItems.map((item) => {
+        return Object.assign({}, item);
+      });
+    }
```
minasmart: I'm pretty sure you could eliminate the branching here.

first off: if you assign `this.cache = []` in the constructor, instead of null, you can just use the code in the `if` part, and clear out the `else` part.

Also, find does a full iteration of the cache for every pass of the line items. Since this is a performance PR, the code below does at most `lineItems.lengthh * 2` iterations:

```javascript
const lineItemCache = this.cache.reduce((acc, item) => {
  acc[item.id] = item;

  return acc;
}, {});

this.cache = lineItems.map((item) => {
  return Object.assign({}, lineItemCache[item.id], item);
});
```

src/views/product.js
```diff
@@ -17,6 +17,7 @@ export default class ProductView extends View {
    * add event listener which triggers resize when the product image is loaded.
    */
   resizeOnLoad() {
+    if (!(this.component.config.product.contents.img || this.component.config.product.contents.imgWithCarousel)) { return; }
```
minasmart: I feel like there should be a linter rule for line length 😄 

This is pretty long.

test/unit/cart.js
```diff
@@ -86,16 +85,29 @@ describe('Cart class', () => {
   describe('fetchData()', () => {
     it('calls fetchRecentCart on client', () => {
       localStorage.setItem('checkoutId', 12345)
-      let fetchCart = sinon.stub(cart.props.client, 'fetchCheckout').returns(Promise.resolve({id: 12345}));
+      let fetchCart = sinon.stub(cart.props.client, 'fetchCheckout').returns(Promise.resolve({id: 12345, lineItems: []}));
```
minasmart: Why is this a let? You're not reassigning.

test/unit/cart.js
```diff
@@ -86,16 +85,29 @@ describe('Cart class', () => {
...
         assert.calledOnce(fetchCart);
         fetchCart.restore();
       });
     });
   });
 
+  describe('fetchMoneyFormat()', () => {
+    it('calls fetchShopInfo on client', () => {
+      localStorage.setItem('checkoutId', 12345)
+      let fetchMoneyFormat = sinon.stub(cart.props.client, 'fetchShopInfo').returns(Promise.resolve({ moneyFormat: '₿{{amount}}'}));
```
minasmart: why let?

test/unit/cart.js
```diff
@@ -146,14 +158,15 @@ describe('Cart class', () => {
         id: 123456,
       }
       cart.view.setFocus = sinon.spy();
-      let addLineItemsStub = sinon.stub(cart.props.client, 'addLineItems').returns(Promise.resolve());
+      let addLineItemsStub = sinon.stub(cart.props.client, 'addLineItems').returns(Promise.resolve({lineItems: [{id: 123, quantity: 1}]}));
```
minasmart: none of these are being reassigned. no let.

src/components/cart.js
lyninx: okay yeah, this reads weirdly. it's because `globalConfig.moneyFormat` takes in the value of `config.moneyFormat` and parses it, but if `config.moneyFormat` is null, it sets a default value. what this logic is doing is: "if moneyFormat was explicitly set, use that value, rather than the this.moneyFormat which was set in the init method". I'll change the init assignment to a different variable to make this clearer.

src/components/cart.js
lyninx: woah. this is so much nicer!! 💯

minasmart: Sometimes I remember how to write code 😛

src/components/cart.js
```diff
@@ -152,13 +155,27 @@ export default class Cart extends Component {
...
    * initializes component by creating model and rendering view.
    * Creates and initalizes toggle component.
    * @param {Object} [data] - data to initialize model with.
    * @return {Promise} promise resolving to instance.
    */
   init(data) {
+    if (!this.config.moneyFormat) {
+      this.fetchMoneyFormat().then((moneyFormat) => {
+        this.moneyFormat = moneyFormat;
+        return moneyFormat;
```
minasmart: Why are you returning when nothing is resolving this promise?

src/components/cart.js
```diff
@@ -152,13 +155,27 @@ export default class Cart extends Component {
...
+      }).catch((err) => {
+        return err;
```
minasmart: I don't think this catch is actually doing anything. Nothing is resolving this promise, and returning from a catch is technically rescuing the promise. Something should handle this error.

src/components/cart.js
```diff
@@ -220,21 +237,54 @@ export default class Cart extends Component {
...
     const newQty = fn(item.quantity);
     return this.props.tracker.trackMethod(this.updateItem.bind(this), 'Update Cart', this.cartItemTrackingInfo(item, newQty))(id, newQty);
   }
 
   /**
+   * set cache using line items.
+   * @param {Array} lineItems - array of GraphModel line item objects.
+   */
+  updateCache(lineItems) {
+    const lineItemCache = this.lineItemCache.reduce((acc, item) => {
```
minasmart: maybe a different name than what you're referencing under `this`. I know this is the name I suggested, but it's confusing when both things are named the same but are objectively different.

src/components/cart.js
lyninx: this is totally a hack. the linter was yelling at me so i just shut it up with this catch LOL
i'll fix it.

.eslintrc
```diff
@@ -2,5 +2,9 @@
   "extends": "plugin:shopify/esnext",
   "env": {
     "browser": true
+  },
+  "rules": {
+  	"promise/always-return": 0,
```
harisaurus: we should be able to disable a particular rule on a line/block instead of disabling it entirely

minasmart: I don't think this rule has merit in most of the project. It's a common use case to use promise resolution for side effects.

lyninx: I figure we don't need those rules in this project anyway, so I'll just leave them disabled globally

harisaurus: 👍

PR #395, opened 2017-07-14
adapt old config to new standards

lyninx:
depends on: https://github.com/Shopify/js-buy-sdk/pull/387

this change allows existing buy buttons, which historically used numerical shopify product IDs, to be transformed into our GraphQL API's product IDs so that they don't fall over once we roll out the updated `buy-button-js` library to existing buy button embeds.

### DIFF
--- a/src/components/product-set.js
+++ b/src/components/product-set.js
@@ -139,7 +139,7 @@ export default class ProductSet extends Component {
       });
     }
     return method.then((collection) => {
-      return Promise.resolve(collection.products);
+      return collection.products;
     });
 
     /* eslint-enable camelcase */
--- a/src/components/product-set.js
+++ b/src/components/product-set.js
@@ -127,31 +127,20 @@ export default class ProductSet extends Component {
    * @param {Object} options - query options for request
    * @return {Promise} promise resolving to collection data.
    */
-  sdkFetch(options = {}) {
-
-    /* eslint-disable camelcase */
-    const queryOptions = Object.assign({}, this.fetchQuery, options);
+  sdkFetch() {
     let method;
+
     if (this.id) {
-      let queryKey;
-      if (isArray(this.id)) {
-        queryKey = 'product_ids';
-      } else {
-        queryKey = 'collection_id';
-        queryOptions.sort_by = 'collection-default';
-      }
-      method = this.props.client.fetchQueryProducts(Object.assign({}, queryOptions, {[queryKey]: this.id}));
+      method = this.props.client.fetchCollectionWithProducts(this.id);
     } else if (this.handle) {
-      method = this.props.client.fetchQueryCollections({handle: this.handle}).then((collections) => {
-        if (collections.length) {
-          const collection = collections[0];
-          this.id = collection.attrs.collection_id;
-          return this.sdkFetch(options);
-        }
-        return Promise.resolve([]);
+      method = this.props.client.fetchCollectionByHandle(this.handle).then((collection) => {
+        this.id = collection.shop.collectionByHandle.id;
+        return this.props.client.fetchCollectionWithProducts(this.id);
       });
     }
-    return method;
+    return method.then((collection) => {
+      return Promise.resolve(collection.products);
+    });
 
     /* eslint-enable camelcase */
   }
--- a/src/components/product-set.js
+++ b/src/components/product-set.js
@@ -134,7 +134,7 @@ export default class ProductSet extends Component {
       method = this.props.client.fetchCollectionWithProducts(this.id);
     } else if (this.handle) {
       method = this.props.client.fetchCollectionByHandle(this.handle).then((collection) => {
-        this.id = collection.shop.collectionByHandle.id;
+        this.id = collection.id;
         return this.props.client.fetchCollectionWithProducts(this.id);
       });
     }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -538,7 +538,7 @@ export default class Product extends Component {
     if (this.id) {
       return this.props.client.fetchProduct(this.id);
     } else if (this.handle) {
-      return this.props.client.fetchQueryProducts({handle: this.handle}).then((products) => products[0]);
+      return this.props.client.fetchProductByHandle(this.handle).then((product) => product);
     }
     return Promise.reject(new Error('SDK Fetch Failed'));
   }
@@ -551,8 +551,8 @@ export default class Product extends Component {
   fetchData() {
     return this.sdkFetch().then((model) => {
       if (model) {
+        this.id = model.id;
         this.handle = model.handle;
-        model.selectedQuantity = 0;
         return model;
       }
       throw new Error('Not Found');
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -52,6 +52,22 @@ function whitelistedProperties(selectorStyles) {
   }, {});
 }
 
+function adaptConfig(config) {
+  if (config.id && !isBase64(config.id)) {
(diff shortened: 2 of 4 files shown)


src/components/product.js
```diff
@@ -52,6 +52,22 @@ function whitelistedProperties(selectorStyles) {
   }, {});
 }
 
+function adaptConfig(config) {
```
minasmart: this isn't doing anything with a config. It's normalizing an ID. Consider a different name.

src/components/product.js
```diff
@@ -52,6 +52,22 @@ function whitelistedProperties(selectorStyles) {
...
+  if (config.id && !isBase64(config.id)) {
+    config.id = btoa(`gid://shopify/Product/${config.id}`);
+  }
+
+  return config;
+}
+
+function isBase64(str) {
+  try {
+    return btoa(atob(str)) === str;
```
minasmart: I'm not sure this is the best way to do this.

```javascript
str = '1234'

btoa(atob(str)) === str
```

The above resolves to true.

It may make sense to have an `id` field, which is deprecated, and a `storefrontId` field, which is the new ID format.

The base64 encoding format doesn't make this a deterministic check if the string is a multiple of 4 characters long.

src/components/product.js
```diff
@@ -461,7 +477,7 @@ export default class Product extends Component {
    * @return {String}
    */
   get onlineStoreURL() {
-    const identifier = this.handle ? this.handle : this.id;
+    const identifier = this.handle ? this.handle : atob(this.id).split('/').pop();
```
minasmart: This is probably an  anti-pattern. We should likely add the online store URL as a field on product. The domain is not always going to be the myshopify domain (which is what's configured here).

src/components/product.js
lyninx: i like this `storefrontId` idea. definitely way nicer than trying to detect what type of ID was provided.

minasmart: also, the only deterministic way 😄

src/components/product.js
swalkinshaw: @minasmart is this valid though? If you have a custom storefront then you might not have online store enabled.

minasmart: you might not, but there are valid use cases to go from the buy button view to the online store view. That's a behaviour that exists today that we can't deprecate.

minasmart: also, we could actually use the same logic we use for blogs to show and hide the value in the API, as opposed to assuming it's always there generating it on the client.

lyninx:
i'll address the onlineStoreURL issue in a separate PR, once we add that to the api #402

src/components/product.js
```diff
@@ -65,7 +75,7 @@ export default class Product extends Component {
    * @param {Object} props - data and utilities passed down from UI instance.
    */
   constructor(config, props) {
-    super(config, props);
+    super(normalizeProductId(config), props);
```
minasmart: I'd use the same pattern we used in `js-buy-sdk`. It's kind of ugly to have a method that mutates its input. Can you receive the config then do `config.id = normalizeProductId(config)`?

src/components/product.js
```diff
@@ -163,7 +173,8 @@ export default class Product extends Component {
    * @return {Object} viewData object.
    */
   get viewData() {
-    return merge(this.model, this.options.viewData, {
+    const model = Object.assign({}, this.model);
+    return merge(model, this.options.viewData, {
```
minasmart: is `merge` different than `Object.assign`?

src/components/product.js
minasmart: Let's add an issue to the storefront API feature board @swalkinshaw created about getting this added to the API so it's not lost.

test/unit/product.js
```diff
@@ -356,7 +356,7 @@ describe('Product class', () => {
 
       it('calls fetchProduct with product id', () => {
```
minasmart: Are you testing with a storefront id, and product id?

src/components/product.js
minasmart: Also, we should probably be preferentially using `storefrontId` everywhere, and deprecate the use of `id`. This can be a different PR though.

src/components/product.js
minasmart: my question is why you don't just do `Object.assign({}, this.model, this.options.viewData, { ... })`

lyninx: oh.... duh. yeah merge isnt requires any more. good call!

test/unit/product.js
lyninx: I'll stick with storefrontId for this test, and make a separate test for the normalization.

src/components/product.js
```diff
@@ -65,6 +75,7 @@ export default class Product extends Component {
    * @param {Object} props - data and utilities passed down from UI instance.
    */
   constructor(config, props) {
+    config.id = normalizeProductId(config).id;
```
minasmart: This assignment is a noop. You're mutating config inside the function. You could do this without the assignment and have the same side effect.

Try to have the method receive data and return data without mutating anything so like:

```javascript
function normalizeProductId(config) {
   if (config.storefrontId) {
     return config.storefrontId;
   } else if (config.id) {
     return = btoa(`gid://shopify/Product/${config.id}`);
   } else {
    throw new Error('No ID was provided');
  }
 }

config.storefrontId = normalizeProductId(config);
```

Mutating inside the function is obfuscating the behaviour you're trying to represent. Also, since the storefront ID is the canonical ID that you're using, storefront ID should be what you're assigning. The ID returned by this is not a database id.

lyninx: wow, yeah not mutating the object in the function seems much cleaner. 💯 
as for `storefrontId` instead of `id`, i'm planning on addressing that by converting the id field across the whole library in a separate pr.

minasmart: 👍

lyninx:
i'm leaving out the error that should be thrown if no ID is provided for now, since there may be early buy-buttons out in the wild which use handles, rather than IDs.

src/components/product-set.js
```diff
@@ -127,31 +127,20 @@ export default class ProductSet extends Component {
...
-        }
-        return Promise.resolve([]);
+      method = this.props.client.fetchCollectionByHandle(this.handle).then((collection) => {
+        this.id = collection.id;
+        return this.props.client.fetchCollectionWithProducts(this.id);
       });
     }
-    return method;
+    return method.then((collection) => {
+      return Promise.resolve(collection.products);
```
lyninx: that it is. will fix.

src/components/product.js
```diff
@@ -65,6 +75,7 @@ export default class Product extends Component {
    * @param {Object} props - data and utilities passed down from UI instance.
    */
   constructor(config, props) {
+    config.id = normalizeProductId(config);
```
lyninx: i'm planning on handling that whole id cleanup as a separate PR

src/components/product.js
```diff
@@ -461,8 +472,7 @@ export default class Product extends Component {
    * @return {String}
    */
   get onlineStoreURL() {
-    const identifier = this.handle ? this.handle : this.id;
-    return `https://${this.props.client.config.domain}/products/${identifier}${this.onlineStoreQueryString}`;
+    return `${this.model.onlineStoreUrl}${this.onlineStoreQueryString}`;
```
lyninx: it adds query params to the url such as the referring channel / domain, as well as which specific variant is chosen

minasmart: ah, cool.

PR #396, opened 2017-07-20
IE10 support

lyninx:
this pr manually adds polyfills from `core-js` in order to support many of the newer language features that i've been using while working on this library.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -17,9 +17,7 @@
         "transform-es2015-parameters",
         "transform-es2015-destructuring",
         "transform-es2015-block-scoping",
-        "transform-es2015-typeof-symbol",
-        "transform-object-assign",
-        "external-helpers"
+        "transform-es2015-typeof-symbol"
       ]
     },
     "development": {
@@ -40,8 +38,7 @@
         "transform-es2015-parameters",
         "transform-es2015-destructuring",
         "transform-es2015-block-scoping",
-        "transform-es2015-typeof-symbol",
-        "transform-object-assign"
+        "transform-es2015-typeof-symbol"
       ]
     }
   }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -72,28 +73,27 @@
...
     "sinon": "1.17.5",
     "testem": "1.9.1",
     "watch": "0.19.1",
     "watchify": "3.7.0",
     "wdio-dot-reporter": "0.0.6",
     "wdio-junit-reporter": "0.1.0",
     "wdio-mocha-framework": "^0.4.0",
-    "webdriverio": "4.2.8"
+    "webdriverio": "4.2.8",
+    "whatwg-fetch": "^2.0.3"
```
minasmart: pin version

package.json
```diff
@@ -72,28 +73,27 @@
     "postcss-cssnext": "2.9.0",
     "postcss-import": "9.0.0",
     "pretender": "1.1.0",
+    "promise-polyfill": "^6.0.2",
```
minasmart: you're not using this.

.babelrc
```diff
@@ -40,9 +38,8 @@
...
         "transform-es2015-destructuring",
         "transform-es2015-block-scoping",
-        "transform-es2015-typeof-symbol",
-        "transform-object-assign"
+        "transform-es2015-typeof-symbol"
       ]
     }
   }
-}
+}
```
minasmart: Is this still being used?

lyninx: yep, since i opted to stick with babel rather than buble due to the fact that we already use `babelify` with `watchify` for development. i couldn't find an equivalent hot-reload for buble.

lyninx: oh snap, i just realized `bubleify` is a thing. do you think it's worth switching to?

PR #412, opened 2017-09-11
Various fixes

harisaurus:
- Update node version in `dev.yml`
- Bump jsbuysdk version for bugfixes
- `clearLineItems` no longer exists. Update to use `removeLineItems` for emptying cart
- Update image getter
  - Use 480 as default if no embed width is provided
  - If % width is provided, use 1000px instead. The % situation is most common for when creating full width embeds using 100%, so 1000px addresses all breakpoints
- Re-add `isSelected` to carousel images to show selected image

### DIFF
--- a/src/components/cart.js
+++ b/src/components/cart.js
@@ -317,10 +317,13 @@ export default class Cart extends Component {
    * Remove all lineItems in the cart
    */
   empty() {
-    return this.model.clearLineItems().then(() => {
+    const lineItemIds = this.model.lineItems ? this.model.lineItems.map((item) => item.id) : [];
+
+    return this.props.client.removeLineItems(this.model.id, lineItemIds).then((checkout) => {
+      this.model = checkout;
       this.view.render();
       this.toggles.forEach((toggle) => toggle.view.render());
-      return;
+      return checkout;
     });
   }
 
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -128,13 +128,20 @@ export default class Product extends Component {
     if (!(this.selectedVariant || this.options.contents.imgWithCarousel)) {
       return null;
     }
-    const imageSize = parseInt(this.options.width, 10) || 480;
+
+    let imageSize;
+    if (this.options.width && this.options.width.slice(-1) === '%') {
+      imageSize = 1000;
+    } else {
+      imageSize = parseInt(this.options.width, 10) || 480;
+    }
+
     let src;
     let id;
 
     const imageOptions = {
       maxWidth: imageSize,
-      maxHeight: imageSize,
+      maxHeight: imageSize * 1.5,
     };
 
     if (this.selectedImage) {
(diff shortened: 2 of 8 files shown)


src/components/cart.js
```diff
@@ -317,7 +317,9 @@ export default class Cart extends Component {
    * Remove all lineItems in the cart
    */
   empty() {
-    return this.model.clearLineItems().then(() => {
+    const lineItemIds = this.model.lineItems.map((item) => item.id);
+
+    return this.props.client.removeLineItems(this.model.id, lineItemIds).then(() => {
```
minasmart: this returns a new checkout instance, which should maybe replace `model`?

src/components/product.js
```diff
@@ -128,13 +128,20 @@ export default class Product extends Component {
     if (!(this.selectedVariant || this.options.contents.imgWithCarousel)) {
       return null;
     }
-    const imageSize = parseInt(this.options.width, 10) || 480;
+
+    let imageSize;
+    if (this.options.width.charAt(this.options.width.length - 1) === '%') {
```
minasmart: This may be more terse with `this.options.width.slice(-1)`

src/components/cart.js
harisaurus: so set `this.model` to this before returning it itself?

src/components/cart.js
```diff
@@ -317,7 +317,9 @@ export default class Cart extends Component {
    * Remove all lineItems in the cart
    */
   empty() {
-    return this.model.clearLineItems().then(() => {
+    const lineItemIds = this.model.lineItems ? this.model.lineItems.map((item) => item.id) : [];
+
+    return this.props.client.removeLineItems(this.model.id, lineItemIds).then(() => {
```
minasmart: How are you setting `this.model` to have empty line items after this returns?

harisaurus:
@minasmart I've made an update to the `empty` function - let me know your thoughts

src/components/cart.js
```diff
@@ -317,11 +317,15 @@ export default class Cart extends Component {
...
-    return this.model.clearLineItems().then(() => {
+    const lineItemIds = this.model.lineItems ? this.model.lineItems.map((item) => item.id) : [];
+
+    this.model = this.props.client.removeLineItems(this.model.id, lineItemIds).then(() => {
       this.view.render();
       this.toggles.forEach((toggle) => toggle.view.render());
       return;
     });
+
+    return this.model;
```
minasmart: so model is a promise now?

minasmart: my thought was more

```javascript
return this.props.client.removeLineItems(this.model.id, lineItemIds).then((updatedCheckout) => {
  this.view.render();
  this.toggles.forEach((toggle) => toggle.view.render());
  this.model = updatedCheckout;

  return this.model;
});
```

That way, anyone calling `empty` will have access to the next cart, and the component itself will update.

harisaurus:
@minasmart updated based on your feedback + our discussion

PR #413, opened 2017-09-14
set storefront variant id when variant id is set + fixes

harisaurus:
- normalize variant ids when present, and use normalized variant id instead of database key.
- a bunch of normalizers
- tiny fixes

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -52,11 +52,11 @@ function whitelistedProperties(selectorStyles) {
   }, {});
 }
 
-function normalizeProductId(config) {
-  if (config.storefrontId) {
-    return config.storefrontId;
-  } else if (config.id) {
-    return btoa(`gid://shopify/Product/${config.id}`);
+function normalizeId(type, config, databaseKey, storefrontKey) {
+  if (config[storefrontKey]) {
+    return config[storefrontKey];
+  } else if (config[databaseKey]) {
+    return btoa(`gid://shopify/${type}/${config[databaseKey]}`);
   } else {
     return null;
   }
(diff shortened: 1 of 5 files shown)


src/components/product.js
```diff
@@ -52,11 +53,11 @@ function whitelistedProperties(selectorStyles) {
   }, {});
 }
 
-function normalizeProductId(config) {
-  if (config.storefrontId) {
-    return config.storefrontId;
-  } else if (config.id) {
-    return btoa(`gid://shopify/Product/${config.id}`);
+function normalizeId(type, config, databaseKey, storefrontKey) {
```
minasmart: Is this still used? OR was it moved to `normalizeConfig`?

src/components/product.js
```diff
@@ -725,11 +727,11 @@ export default class Product extends Component {
    */
   setDefaultVariant(model) {
     let selectedVariant;
+    if (this.defaultStorefrontVariantId) {
+      selectedVariant = model.variants.find((variant) => btoa(`gid://shopify/ProductVariant/${variant.id}`) === this.defaultStorefrontVariantId);
```
minasmart: maybe export `normalizeId` from `normalizeConfig` too, so you can share logic

test/unit/product.js
```diff
@@ -430,29 +430,41 @@ describe('Product class', () => {
...
     });
 
     it('calls init if variant ID updated', () => {
       product.updateConfig({variantId: 7777});
       assert.calledOnce(initSpy);
-      assert.equal(product.defaultVariantId, 7777);
+      assert.equal(product.defaultStorefrontVariantId, 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC83Nzc3');
+    });
+
+    it('calls init if storefront variant ID updated', () => {
```
minasmart: so glad to see this.

src/components/product.js
harisaurus: `normalizeId` is a little more complex than this scenario - it manages variantId when on a config object, whereas this context has no `config`.

might need to rework the normalizers to be a little more reusable.

PR #416, opened 2017-09-29
Switch to the js buy beta

minasmart:
all the tests pass. nothing is skipped. one weird thing, there's a test that somehow leaves a modal open. will mark in comments. @harisaurus can you really top hat this hard core?


test/unit/product.js
```diff
@@ -406,7 +419,7 @@ describe('Product class', () => {
       assert.calledWith(cartUpdateConfigSpy, newConfig);
     });
 
-    it.skip('calls updateConfig on modal if modal exists', () => {
+    it('calls updateConfig on modal if modal exists', () => {
```
minasmart: @harisaurus this somehow leaves a modal open. after it runs. If you debug this test, you'll see it close its modal, but something in here is doing something async, and after the test finishes, a modal is re-opened. If you skip this test, it's no longer an issue. Please help.

PR #417, opened 2017-10-02
prevent events from binding again on re-inits

harisaurus:
Anytime the config is changed for an embed, `init` is called. This calls `delegateEvents` which would create event listeners for the required elements. This was an issue as multiple listeners were attached to the same elements over and over anytime the config was updated, which would cause crazy behaviour in the Buy Button app.

There should ideally be an `init` and a similar `update` method, the latter not rebinding events, but that would require a bunch of rework. Adding in the simpler, but not really elegant, solution as a flag that is checked before the events are bound.

Ping me for 🎩 -ing instructions to test buybuttonjs with buybutton.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -13,6 +13,7 @@ export default class View {
     this.iframe = null;
     this.node = this.component.node;
     this.template = new Template(this.component.options.templates, this.component.options.contents, this.component.options.order);
+    this.eventsBound = false;
   }
 
   init() {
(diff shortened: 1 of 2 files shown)


src/view.js
```diff
@@ -53,7 +54,13 @@ export default class View {
    * delegates DOM events to event listeners.
    */
   delegateEvents() {
+
```
minasmart: 🔥

PR #425, opened 2017-12-04
add srcLarge for default horizontal layouts and update example

harisaurus:
- Updated the `index.example.html` starter file
- Added a new key to the `image` object for `srcLarge`. We need to use this for horizontal layouts and primary images in the modal. The current `src` url image has a max height/width of 480 which is not sufficient.

`srcLarge` isn't the best name. if you've got a better suggestion that'd be sweet. also, ideally the `src` key should just have various sizes listed under it, but this will require significant change. we can perhaps focus on that at a later time.

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -117,6 +117,9 @@ export default class Product extends Component {
    * @return {Object} image object.
    */
   get image() {
+    const DEFAULT_IMAGE_SIZE = 480;
+    const MODAL_IMAGE_SIZE = 550;
+
     if (!(this.selectedVariant || this.options.contents.imgWithCarousel)) {
       return null;
     }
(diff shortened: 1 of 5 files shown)


src/components/product.js
```diff
@@ -128,22 +128,31 @@ export default class Product extends Component {
       imageSize = parseInt(this.options.width, 10) || 480;
     }
 
-    let src;
     let id;
+    let src;
+    let srcLarge;
+    const MODAL_IMAGE_SIZE = 550;
```
swalkinshaw: maybe define this at the top of the function?

Would also be nice to define that `480` default from above as a constant along with it.

harisaurus:
Thanks, @swalkinshaw for the feedback. Made the changes. 👍

PR #430, opened 2017-12-12
always use a checkout for the checkout button destination.

minasmart:
This also properly opens the checkout window synchronously when the button destination is `checkout`, and is compatible with non-straight-to-checkout buttons on the page

### DIFF
--- a/src/components/cart.js
+++ b/src/components/cart.js
@@ -296,11 +296,13 @@ export default class Cart extends Component {
    * @param {Object} variant - variant object.
    * @param {Number} [quantity=1] - quantity to be added.
    */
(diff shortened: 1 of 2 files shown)


src/components/cart.js
```diff
@@ -296,11 +296,13 @@ export default class Cart extends Component {
    * @param {Object} variant - variant object.
    * @param {Number} [quantity=1] - quantity to be added.
    */
-  addVariantToCart(variant, quantity = 1) {
+  addVariantToCart(variant, quantity = 1, openCart = true) {
     if (quantity <= 0) {
       return null;
     }
-    this.open();
+    if (openCart) {
```
harisaurus: oh this is very nice!

PR #442, opened 2018-02-27
Fix behaviour for invalid localStorage key

StefanSlehta:
Bump the `shopify-buy` version and `fetchData()` to use the newly added behaviour. 
The code assumed that if `checkoutId` is defined in local storage, it's also valid. This isn't always the case and the behaviour is accounted for in this PR.

### DIFF
--- a/src/components/cart.js
+++ b/src/components/cart.js
@@ -132,6 +132,14 @@ export default class Cart extends Component {
     }
   }
 
+  createCheckout() {
+    return this.props.client.checkout.create().then((checkout) => {
+      localStorage.setItem('checkoutId', checkout.id);
+      this.model = checkout;
+      return checkout;
+    });
+  }
+
   /**
    * get model data either by calling client.createCart or loading from localStorage.
    * @return {Promise} promise resolving to cart instance.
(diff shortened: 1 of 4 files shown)


src/components/cart.js
```diff
@@ -132,6 +132,14 @@ export default class Cart extends Component {
     }
   }
 
+  createCheckout() {
```
minasmart: Use other methods to compare, but is this by default private, or do you need a jsdoc comment?

test/unit/cart.js
```diff
@@ -91,6 +91,19 @@ describe('Cart class', () => {
     });
   });
 
+  describe('fetchData()', () => {
+    it('calls createCart on client if localStorage invalid', () => {
+      localStorage.setItem('checkoutId', 1);
+      const createCheckout = sinon.stub(cart.props.client.checkout, 'create').returns(Promise.resolve({id: 12345, lineItems: []}));
```
minasmart: This isn't testing what you think it's checking. You should be stubbing `checkout.fetch` to return a `Promise.reject`, then ensure that `createCheckout` is called. We want to see that when fetch fails, create is called.

src/components/cart.js
StefanSlehta: Added a jsdoc for this

PR #444, opened 2018-03-02
Fallback strategy when no variant or product image is defined

StefanSlehta:
Latest storefront failure was due to `selectedVariant.image` being undefined. This PR should address the issue by using the `NO_IMG_URL` form cdn, or the first product image, if it exists

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -20,6 +20,8 @@ function isMedia(key) {
   return key.charAt(0) === '@';
 }
 
+const NO_IMG_URL = 'https://sdks.shopifycdn.com/buy-button/latest/no-image.jpg';
+
 const ENTER_KEY = 13;
 
 const propertiesWhitelist = [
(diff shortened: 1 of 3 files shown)


src/components/product.js
```diff
@@ -141,6 +143,14 @@ export default class Product extends Component {
       id = this.selectedImage.id;
       src = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptionsLarge);
+    } else if (this.selectedVariant.image == null && this.model.images[0] == null) {
+      id = null;
+      src = NO_IMG_URL;
+      srcLarge = NO_IMG_URL;
```
minasmart: You can use the `imageForSize` helper here too I think.

test/unit/product.js
```diff
@@ -10,6 +10,7 @@ import shopFixture from '../fixtures/shop-info';
 import productFixture from '../fixtures/product-fixture';
 
 const rootImageURI = 'https://cdn.shopify.com/s/';
+const NO_IMG_URL = 'https://sdks.shopifycdn.com/buy-button/latest/no-image.jpg';
```
minasmart: Don't duplicate this. export it from the product component

src/components/product.js
StefanSlehta: I've tried putting this through `imageForSize` but I'm getting `403`s as a response with (the link that's generated)[https://sdks.shopifycdn.com/buy-button/latest/no-image_550x825.jpg]

PR #445, opened 2018-03-06
Fix describe options to only ever select 1 option

StefanSlehta:
`describeOptions` would sometimes select multiple options if we have duplicated values in different dropdown lists. This PR should fix it

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -396,7 +396,7 @@ export default class Product extends Component {
         values: option.values.map((value) => {
           return {
             name: value.value,
-            selected: Object.values(this.selectedOptions).some((selectedOption) => {
+            selected: Object.values([this.selectedOptions[option.name]]).some((selectedOption) => {
               return selectedOption === value.value;
             }),
           };
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -396,9 +396,7 @@ export default class Product extends Component {
         values: option.values.map((value) => {
           return {
             name: value.value,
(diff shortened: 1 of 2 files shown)


src/components/product.js
```diff
@@ -396,7 +396,7 @@ export default class Product extends Component {
         values: option.values.map((value) => {
           return {
             name: value.value,
-            selected: Object.values(this.selectedOptions).some((selectedOption) => {
+            selected: Object.values([this.selectedOptions[option.name]]).some((selectedOption) => {
```
minasmart: Object.values on an array returns that array. I think this is calling `some` on an array of one item. I think n this call is equivalent to `this.selectedOptions[option.name] === value.value`

test/unit/product.js
```diff
@@ -295,6 +295,20 @@ describe('Product class', () => {
...
+      expectedArray[1].values.push({name: 'something', selected: false});
+
+      return product.init(testProductCopy).then(() => {
+        product.model.options[0].values.push({value: 'something'});
+        product.model.options[1].values.push({value: 'something'});
+
+        product.updateVariant('Print', 'something');
+        assert.deepEqual(product.decoratedOptions, expectedArray);
+      });
+    });
```
minasmart: Did this test fail before the change?

StefanSlehta: Yes.

minasmart: 👍🏻

StefanSlehta:
I also regression tested with a regular product that has `size/color` variants

PR #448, opened 2018-03-07
Remove NO_IMG_URL as a fallback image and render nothing

StefanSlehta:
Closes [#2334](https://github.com/Shopify/buy-button/issues/2334)


StefanSlehta:
I'll look into removing the `wrapperClass` in case there's no image

PR #455, opened 2018-03-13
selectedQuantity value is not stored in in the model

SpankyWorks:
I noticed this when using the CDN delivered `buybutton.js` from the url https://sdks.shopifycdn.com/buy-button/latest/buybutton.js :

Updating the quantity and clicking the "Add To Cart" button does not add the `selectedQuantity` of the product to the cart because it's looking at `this.model.selectedQuantity` when that value is actually stored at `this.selectedQuantity`.

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -428,7 +428,7 @@ export default class Product extends Component {
     return {
       id: variant.id,
       name: variant.productTitle,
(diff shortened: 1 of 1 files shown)


minasmart:
Missing tests. Updated in #458. 

Thanks for your help!!

PR #467, opened 2018-04-09
Create new checkout if the locally stored one is completed

StefanSlehta:
Makes a call to `createCheckout` if the `completedAt` exists on the current one. After the checkout is completed, it'll just call `location.reload` and the cart will be reset to a fresh state.


test/unit/cart.js
```diff
@@ -132,9 +132,7 @@ describe('Cart class', () => {
         fetchCart.restore();
       });
     });
-  });
 
-  describe('fetchData()', () => {
```
minasmart: why do we lose this describe?

test/unit/cart.js
```diff
@@ -147,6 +145,19 @@ describe('Cart class', () => {
...
       });
     });
+
+    it('calls createCart on client if checkout is completed', () => {
+      localStorage.setItem('checkoutId', 123);
+      const fetchCart = sinon.stub(cart.props.client.checkout, 'fetch').returns(Promise.resolve({ completedAt: "04-12-2018", lineItems: []}));
+      const createCheckout = sinon.stub(cart.props.client.checkout, 'create').returns(Promise.resolve({id: 1111, lineItems: []}));
+
+      return cart.fetchData().then((data) => {
+        assert.deepEqual(data, {id: 1111, lineItems: []});
```
minasmart: stick your static ids in vars

minasmart: or even the whole stub value.

test/unit/cart.js
StefanSlehta: This is a duplicate , I think I mistakenly added it in one of the previous PRs, these tests are still under the `fetchData()`

PR #471, opened 2018-04-16
Make the checkout local storage key unique per shopify buy client

minasmart:
This stops pages with multiple embeds from different shops and different
channels from clobbering each others cart IDs

### DIFF
--- a/src/buybutton.js
+++ b/src/buybutton.js
@@ -15,13 +15,13 @@ ShopifyBuy.UI = window.ShopifyBuy.UI || {
   domains: {},
 
   init(client, integrations = {}, styleOverrides) {
-    const domain = client.shop.fetchInfo().then((res) => {
-      return res.attrs.primaryDomain.attrs.host.value;
-    });
-    if (!this.domains[domain]) {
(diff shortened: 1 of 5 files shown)


src/components/cart.js
```diff
@@ -119,6 +119,10 @@ export default class Cart extends Component {
     return this.isVisible ? 'is-active' : '';
   }
 
+  get localStorageCheckoutKey() {
+    return `${this.props.client.config.storefrontAccessToken}.${this.props.client.config.domain}.checkoutId`;
```
harisaurus: 👌

src/buybutton.js
```diff
@@ -15,13 +15,13 @@ ShopifyBuy.UI = window.ShopifyBuy.UI || {
   domains: {},
 
   init(client, integrations = {}, styleOverrides) {
-    const domain = client.shop.fetchInfo().then((res) => {
```
minasmart: @harisaurus turns out this was a long standing bug breaking any multi-shop set ups.

minasmart:
__NOTE__: This will not fix styling issues with having two cart drawers. That work requires custom style overrides on the client.

PR #489, opened 2018-07-30
Empty cart when direct checkout button is clicked

spencercanner:
#### What are you trying to accomplish with this PR?
Currently, the direct checkout button adds an item to the cart, but does not clear the cart before doing so. This can lead to buyers proceeding with a checkout that contains multiple items, contrary to what is found in the help docs: `Direct checkout brings the customer to a checkout page, where they can purchase the product. If a customer chooses this option, then they aren't able to add additional products to their order.`

old-cart

Ideally, clicking the direct checkout button should always lead to a cart which only contains the variant selected. 

#### How does it do it?
The cart gets emptied before adding the specific variant to the cart. 

new-cart

It should be noted that this does not resolve the issue of the cart toggle appearing when the button is clicked. This only addresses the problem of a direct checkout button resulting in a multi-item cart. 

I also considered running `atob` on `this.selectedVariant.id` to decode the guid, and use its known structure to extract the variant id. This could then be used to build a cart permalink, which would be injected directly into the checkout window. This approach would also solve the issue of the cart toggle appearing, however I feel that it's cleaner to go through the js-buy-sdk.

### DIFF
--- a/test/unit/product.js
+++ b/test/unit/product.js
@@ -802,5 +802,31 @@ describe('Product class', () => {
         title: "sloth / small"
       }, 1111);
     });
+
+    it('empty cart and add variant to cart are called when destination is checkout', async () => {
+      product.config.product.buttonDestination = 'checkout';
+      product.config.cart.popup = true;
+      sinon.stub(product.cart, 'close');
+      sinon.stub(window, 'open').returns({location: ''});
+      
+      const emptyCart = sinon.stub(product.cart, 'empty').returns(Promise.resolve());
+      const addToCart = sinon.stub(product.cart, 'addVariantToCart').returns(Promise.resolve({webUrl: 'test'}));
+      const evt = new Event('click shopify-buy__btn--parent');
+      const target = 'shopify-buy__btn--parent';
+
+      await product.onButtonClick(evt, target);
+
+      assert.calledOnce(emptyCart);
+      assert.calledOnce(addToCart);
+      assert.calledWith(addToCart, {
+        available: true,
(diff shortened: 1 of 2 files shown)


spencercanner:
@minasmart I added a test for the behaviour in this repo. Does this PR need to be merged before `buy-button-storefront` changes can be done?

test/unit/product.js
```diff
@@ -802,5 +802,31 @@ describe('Product class', () => {
         title: "sloth / small"
       }, 1111);
     });
+
+    it('empty cart and add variant to cart are called when destination is checkout', async () => {
+      product.config.product.buttonDestination = 'checkout';
+      product.config.cart.popup = true;
+      sinon.stub(product.cart, 'close');
```
minasmart: does this stub need to exist?

minasmart:
@spencercanner changes look good.

The process would be:
1. Merge this PR
2. In your local copy of this repo, from master, `git pull && npm version minor && git push && git push --tags`
3. Deploy to NPM through https://shipit.shopify.io/shopify/buy-button-js/production
4. New branch on `buy-button-storefront`
5. Bump version of `buy-button-js` in `buy-button-storefront/package.json`, run tests, possibly fix.

test/unit/product.js
spencercanner: I'm going to leave it in, but actually assert that it was called.

minasmart: I guess my only question is, is this testing behaviour, or testing a known implementation?

spencercanner:
Closing, since this issue is being resolved in #490

Issue #492, opened 2018-08-10
invalid aria src/templates/modal.js


limograf:
role cannot have a state value
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques
<span aria-role="hidden">&times;</span>

should be

<span aria-hidden="true">&times;</span>

PR #493, opened 2018-08-10
Fixed invalid aria-role #492

limograf:
#492 role cannot have a state value
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques


gabrielmaldi:
This should be merged.

This is my current workaround:

```js
events: {
  beforeInit: function (component) {
    component.view.template.templates.contents = component.view.template.templates.contents.replace(
      'aria-role="hidden"',
      'aria-hidden="true"'
    );
  }
}
```

PR #494, opened 2018-08-15
Add more tracking for the product component

PhilibertDugas:
## What are you trying to accomplish with this PR?
Looking to add a bit more attributes to the current tracking to be able to calculate click-rate

For this we need to have amounts of click + amounts of views per destination

## How is this accomplished?
- Add visibility around the `buttonDestination` attribute
- Track the event of direct checkout being clicked

With a custom tracker that logs the events, I was able to test locally
image

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -412,7 +412,7 @@ export default class Product extends Component {
     };
 
     if (this.selectedVariant) {
-      return Object.assign({}, info, {
+      Object.assign(info, {
         id: this.id,
         name: this.selectedVariant.productTitle,
         sku: null,
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -413,9 +413,12 @@ export default class Product extends Component {
         name: this.selectedVariant.productTitle,
         sku: null,
         price: this.selectedVariant.price,
+        destination: this.options.buttonDestination,
       };
     } else {
-      return {};
+      return {
+        destination: this.options.buttonDestination,
+      };
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/components/product.js
```diff
@@ -407,16 +407,20 @@ export default class Product extends Component {
...
    */
   get trackingInfo() {
+    const info = {
+      destination: this.options.buttonDestination,
+    };
+
     if (this.selectedVariant) {
-      return {
+      return Object.assign({}, info, {
         id: this.id,
```
minasmart: Why include the empty object? If you just assign to info, there’s no need to return inside the if and this method just has one return stanza

test/unit/product.js
```diff
@@ -50,7 +50,10 @@ describe('Product class', () => {
           return function () {
             fn(...arguments);
           }
-        }
+        },
+        track: (eventName, properties) => {
```
minasmart: Shouldn’t tests capture this instead of just blind stubbing it?

src/components/product.js
saulecabrera: @minasmart valid observation; it's a matter of coding styles, but mostly best practices I guess, not mutating objects after creation brings a series of benefits, mostly related to avoiding unclear states.

minasmart:
I would agree if your identity mattered, but given the immediate return there’s no misalignment between identity and state. Additionally, eliminating return path branches isn’t a style issue as much as it is a readability issue. A single point of return is way simpler to reason about.

— Mina

PhilibertDugas:
Updated the PR with the proposed changes

test/unit/product.js
```diff
@@ -843,6 +844,8 @@ describe('Product class', () => {
       });
 
       product.onButtonClick(evt, target);
+
+      assert.calledWith(product.props.tracker.track, 'Direct Checkout', {})
```
minasmart: 👍

PR #495, opened 2018-08-16
Fix direct checkout quantity

spencercanner:
In #490, the direct checkout behaviour was updated to create a new cart for checkout, and add the selected variant to it. However, a quantity of 1 was always being added, regardless of the quantity input. 
This PR adds the expected quantity to the cart. 

Before:
bad-buy-now

After:
good-buy-now

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -593,7 +593,7 @@ export default class Product extends Component {
       }
 
       this.props.client.checkout.create().then((checkout) => {
-        const lineItem = {variantId: this.selectedVariant.id, quantity: 1};
+        const lineItem = {variantId: this.selectedVariant.id, quantity: this.selectedQuantity};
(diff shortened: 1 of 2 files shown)


src/components/product.js
```diff
@@ -593,7 +593,7 @@ export default class Product extends Component {
       }
 
       this.props.client.checkout.create().then((checkout) => {
-        const lineItem = {variantId: this.selectedVariant.id, quantity: 1};
+        const lineItem = {variantId: this.selectedVariant.id, quantity: this.selectedQuantity};
```
minasmart: is `this.selectedQuantity` defaulted anywhere?

spencercanner: yup, it's done in the product's constructor: https://github.com/Shopify/buy-button-js/blob/c5b9c87f64f87cdcbcc7fa2d12c5f38cebb5afe6/src/components/product.js#L79

minasmart: 👍

PR #498, opened 2018-08-22
Use the globalConfig's moneyFormat for the cart

spencercanner:
The `moneyFormat` passed in via `createComponent` is available on the `globalConfig` not on `config`, which is causing the cart's money format to always be fetched from the js-buy-sdk. 
To determine whether we need to be fetching it from the sdk, we should be checking `this.moneyFormat`, which is being set in the constructor based off of the `globalConfig` value.

### DIFF
--- a/test/unit/cart.js
+++ b/test/unit/cart.js
@@ -28,7 +28,8 @@ describe('Cart class', () => {
             fn(...arguments);
           }
         }
-      }
+      },
+      fetchMoneyFormat: function () { return Promise.resolve('test') }
     });
   });
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


StefanSlehta:
Can we also add a test for this? There's already a [fetchMoneyFormat() suite](https://github.com/Shopify/buy-button-js/blob/6759ea67c0dd3b6eb8419765a740959aa4f842ee/test/unit/cart.js#L150) but nothing tests for when it should be called.

test/unit/cart.js
```diff
@@ -245,4 +246,26 @@ describe('Cart class', () => {
...
+      cart.moneyFormat = null;
+      const fetchMoneyFormatStub = sinon.stub(cart.props, 'fetchMoneyFormat').returns(Promise.resolve());
+
+      cart.init().then(() => {
+        assert.calledOnce(fetchMoneyFormatStub);
+        fetchMoneyFormatStub.restore();
+      });
+    });
+
+    it('calls does not call fetchMoneyFormat when moneyFormat has been set', () => {
```
StefanSlehta: `does not call fetchMoneyFormat when moneyFormat has been set`

PR #503, opened 2018-09-12
Re-add html escaping to cart total and line-item prices

melissaluu:
In this PR, HTML escaping will be re-added to the `pricing` values within `cart` and `line-item` templates. This fixes [#2455](https://github.com/Shopify/buy-button/issues/2455).


clayton-shopify:
For reference, this reverts #451.

PR #505, opened 2018-09-24
Add custom outerHeight() to Product component to always return clientHeight

melissaluu:
This PR updates the `product` component to always return the `clientHeight` as the outer height of the product container.  This outer height is used to set the height of the `iFrame` container.

Fixes: [#2497](https://github.com/Shopify/buy-button/issues/2497), [#2452](https://github.com/Shopify/buy-button/issues/2452), [#2366](https://github.com/Shopify/buy-button/issues/2366)

### DIFF
--- a/src/views/product.js
+++ b/src/views/product.js
@@ -13,6 +13,10 @@ export default class ProductView extends View {
     return true;
   }
 
+  get outerHeight() {
+    return `${this.wrapper.clientHeight}px`;
+  }
+
   /**
    * add event listener which triggers resize when the product image is loaded.
    */
--- a/test/unit/product.js
+++ b/test/unit/product.js
@@ -850,4 +850,19 @@ describe('Product class', () => {
       assert.calledWith(product.props.tracker.track, 'Direct Checkout', {})
     });
   });
+
+  describe('get view.outerHeight', () => {
+    beforeEach(() => {
+      return product.init(testProductCopy).then((product) => {
+        product.cart.model.lineItems = [];
+        product.cart.props.client = product.props.client;
+        return Promise.resolve();
+      });
+    });
+
+    it('returns the wrapper\'s client height in px', () => {
+      const wrapperHeight = product.view.wrapper.clientHeight;
+      assert.equal(product.view.outerHeight, `${wrapperHeight}px`);
+    });
+  });
 });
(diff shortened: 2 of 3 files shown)


src/views/product.js
```diff
@@ -13,6 +13,10 @@ export default class ProductView extends View {
     return true;
   }
 
+  get outerHeight() {
+    return `${this.wrapper.clientHeight}px`;
+  }
+
```
minasmart: is there a usage of this prop anywhere in the project? And should there be a test?

melissaluu: It's used in `resizeY()` within `view.js` during the `render` process. https://github.com/Shopify/buy-button-js/blob/ba89d7fa2f4944da7f388a9b6d632d764b376e5f/src/view.js#L270

In terms of tests, it seems custom `view` classes aren't really tested (I was following how the `toggle` component's custom `outerHeight()` was implemented and that didn't have test coverage either).
https://github.com/Shopify/buy-button-js/blob/ba89d7fa2f4944da7f388a9b6d632d764b376e5f/src/views/toggle.js#L22

minasmart: huh. I'm curious how the usage in `view.js` works if this hasn't been merged yet.

Also, I'm going to question two things:

If this is just an alias for `this.wrapper.clientHeight` why doesn't the caller just access `this.wrapper.clientHeight`. Adding it as an alias makes it a public method that needs to be documented and supported.

Second: Without a clear expectation in a test for this alias to exist _or_ for the behaviour that it enables, it's quite hard to understand why we would need it.

melissaluu: Just added a test for product view `outerHeight` property 😄 

>I'm also curious now though, is the computed style ever worthwhile? What do you miss out on by just defaulting to this.wrapper.clientHeight everywhere?

Good question - TBH, I'm not entirely sure how defaulting to `this.wrapper.clientHeight` everywhere would affect the individual components. @harisaurus - do you have any insight on this?  I think for now, we can use the override solution to fix the immediate FF issue and address this in a future PR  since updating the `view.outerHeight` will need extensive testing.

melissaluu:
@minasmart @spencercanner - would you be able to take a look at this PR again? I've added a few additional tests to better cover the resizing of the `iFrame`

PR #508, opened 2018-10-11
Remove line items with null variants

### LINKED ISSUE #488: BuyButton fails to load if the cart previously had items in it
Hello,

My partner uses the buy button on her Wordpress site. Recently we went on the site and noticed none of the buy button bits were loading.

A quick check in the console shows the error:

```
TypeError: Cannot read property 'image' of null
    at e.imageForLineItem (buy-button-storefront.min.js:8)
    at buy-button-storefront.min.js:8
    at Array.reduce (<anonymous>)
    at e.get (buy-button-storefront.min.js:8)
    at e.get (buy-button-storefront.min.js:8)
    at e.t.render (buy-button-storefront.min.js:7)
    at e.render (buy-button-storefront.min.js:8)
    at buy-button-storefront.min.js:7
```

The error occurs as the result from the graphql query doesn't return variants for the item in the cart, and subsequently blows up. An immediate fix for this could be to change line 132 in src/components/cart.js from

`if (lineItem.variant.image) {`

to

`if (lineItem.variant && lineItem.variant.image) {`

However I haven't tested it, and wouldn't be sure this fully fixes the issue, for a start the cart would probably still struggle to render the item.

After a bit more delving into how the cart is restored, I tried deleting the local storage, which on the next reload seems to fix the issue.

Not sure if this is a backend issue; base64 decoding the local storage value shows it contains some sort of id e.g. gid://shopify/Checkout/<ID>, or something else?

Obviously it's not an easy fix for users to do, so a potential fix could be to purge the local storage if the graphql query comes back with partial/invalid data?

Hopefully I've included enough detail, if you have any questions, let me know. If you think this sounds sensible I can always attempt to whip up a PR with the suggested changes.

Thanks,

Vince

spencercanner:
Closes #488 
Currently storefront api returns line items containing variants which have been deleted. This PR adds sanitization to the checkout returned from js-buy-sdk, which will remove any line items that have null variants. These line items must be removed via the js-buy-sdk, rather than just being hidden from the ui, since they're included in the checkout's subtotal. 

To test:
1. Create buy buttons for multiple products with variants, with a button destination of cart
2. After embedding them in a page, add some variants to your cart
3. In admin, delete a variant that's in your cart 
4. Refresh the buy button page
5. Verify that the cart no longer contains the deleted variant, and the subtotal is correct
6. You can also try removing an entire product, and verify that the checkout's line item is removed

### DIFF
--- a/test/unit/cart.js
+++ b/test/unit/cart.js
@@ -146,6 +146,36 @@ describe('Cart class', () => {
         assert.equal(localStorage.getItem(cart.localStorageCheckoutKey), checkout.id);
       });
     });
+
+    it('calls sanitizeCheckout then updateCache with the new checkout', () => {
+      localStorage.setItem(cart.localStorageCheckoutKey, 123);
+      
+      const checkout = {id: 1111, lineItems: [
+        {id: 1112, variant: null},
+        {id: 1113, variant: {id: 1114}},
+        {id: 1115, variant: null},
+      ]};
+      const sanitizedCheckout = {id: 1111, lineItems: [
+        {id: 1113, variant: {id: 1114}},
+      ]};
+      
+      const fetchCart = sinon.stub(cart.props.client.checkout, 'fetch').resolves(checkout);
+      const sanitizeCheckout = sinon.stub(cart, 'sanitizeCheckout').resolves(sanitizedCheckout);
+      const updateCache = sinon.stub(cart, 'updateCache').resolves();
+
+      return cart.fetchData().then((data) => {
+        assert.deepEqual(data, sanitizedCheckout);
+        assert.calledOnce(fetchCart);
+        assert.calledOnce(sanitizeCheckout);
+        assert.calledOnce(updateCache);
+        assert.calledWith(sanitizeCheckout, checkout);
+        assert.calledWith(updateCache, sanitizedCheckout.lineItems);
+
+        fetchCart.restore();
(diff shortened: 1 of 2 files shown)


test/unit/cart.js
```diff
@@ -161,6 +191,28 @@ describe('Cart class', () => {
     });
   });
 
+  describe('sanitizeCheckout()', () => {
+    it('calls removeLineItems for all line items with deleted variants', () => {
```
minasmart: This is a bit of a nitpick, but I'm wondering if this and the test above are coupled too tightly to the implementation. Ideally, we just test behaviours. ex: We want to ensure that line items with null variants are removed, and you do do that, but checking the calls makes this very brittle if the implementation of how that's done ever changes. If the implementation changed, we'd still want to ensure the behaviour, but the test would still fail.

I'm curious what your thoughts are, and if you see a strong value in testing the implementation, if it'd be worthwhile to break that out into a separate test so the behavioural tests still work as insurance and are fully decoupled from internals.

spencercanner: Since these tests require some stubbing, I think the implementation is always going to be somewhat implicitly tested based on what functions are being stubbed. I'm explicitly checking the calls for completeness, since we already have the stubs. If the implementation were to change they would most likely need to update the stubs anyways, even if the test was only checking for the behaviour.

PR #528, opened 2018-10-31
added unit tests for ui.js

jameskim08:
created unit tests for ui.js

### DIFF
--- a/test/unit/ui.js
+++ b/test/unit/ui.js
@@ -141,14 +141,6 @@ describe('ui class', () => {
       });
     });
 
-    it('passes config to constructor', () => {
-      productConfig.node = null;
-      return ui.createComponent('product', productConfig).then(() => {
-        assert.equal(null, ui.components.product[0].config.node);
-        ui.destroyComponent('product', ui.components.product[0].model.id);
-      });
-    });
-
     it('binds escape key on iframe if it exists', () => {
       productConfig.iframe = true;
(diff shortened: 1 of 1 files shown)


test/unit/ui.js
```diff
@@ -87,38 +132,206 @@ describe('ui class', () => {
...
+      ui.trackComponent('productSet', productSet);
+      assert.callCount(trackComponentStub, 5);
+    });
+
+    it('calls trackComponent with the given component type', () => {
+      const component = {
+        trackingInfo: 'test',
+      };
+
+      ui.trackComponent('product', component);
```
spencercanner: Do you think we need to explicitly test each component type, or would it suffice to just test a single generic component type?

test/unit/ui.js
```diff
@@ -87,38 +132,206 @@ describe('ui class', () => {
...
+    it('creates new component of type with tracker attached', () => {
       return ui.createComponent('product', productConfig).then(() => {
         assert.equal(1, ui.components.product.length);
+        assert.calledOnce(initStub);
+        assert.calledOnce(trackStub);
         ui.destroyComponent('product', ui.components.product[0].model.id);
       });
     });
 
     it('passes config to constructor', () => {
```
spencercanner: What is this testing? The title doesn't really seem related to what's going on inside the test 😕

PR #532, opened 2018-11-01
Fixed console errors that appear while running tests

jameskim08:
Removed deprecated use of sinon.spy() with three parameters
Stubbed init to avoid fetch calls on the client

### DIFF
--- a/test/unit/cart.js
+++ b/test/unit/cart.js
@@ -303,7 +303,7 @@ describe('Cart class', () => {
     let fetchMoneyFormatStub;
     beforeEach(() => {
       superInitStub = sinon.stub(Component.prototype, 'init').resolves({model: {lineItems: [{id: 123, quantity: 5}]}});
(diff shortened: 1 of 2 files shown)


test/unit/cart.js
```diff
@@ -300,23 +299,29 @@ describe('Cart class', () => {
   });
 
   describe('init', () => {
+    let superInitStub;
+    let fetchMoneyFormatStub;
+    beforeEach(() => {
+      superInitStub = sinon.stub(Component.prototype, 'init').resolves({model: {lineItems: [{id: 123, quantity: 5}]}});
+      fetchMoneyFormatStub = sinon.stub(cart, 'fetchMoneyFormat').returns(Promise.resolve('test'));
```
melissaluu: Let's be consistent and use the `resolves()` function instead of the `returns(Promise....)`. 

Also, if we're not testing the resolved promise's return, let's leave it as an empty promise return.

PR #534, opened 2018-11-09
Updated constructor tests for ui.js

jameskim08:
- fixed eslint errors
- separated constructor test from the other tests
- created a beforeEach that handles instantiating a UI object for all constructor tests

### DIFF
--- a/test/unit/ui.js
+++ b/test/unit/ui.js
@@ -1,124 +1,210 @@
 import ShopifyBuy from '../../src/buybutton';
 import UI from '../../src/ui';
 import Product from '../../src/components/product';
+import Tracker from '../../src/utils/track';
+import ProductSet from '../../src/components/product-set';
 import Cart from '../../src/components/cart';
+import CartToggle from '../../src/components/toggle';
 import shopFixture from '../fixtures/shop-info';
+
+const DATA_ATTRIBUTE = 'data-shopify-buy-ui';
 
 describe('ui class', () => {
   let ui;
   let script;
 
   let client;
+  let integrations;
   let config;
-  const productConfig = {
-    id: 123,
-    options: {}
-  };
 
   beforeEach(() => {
+    integrations = {
+      errorReporter: {notifyException: sinon.spy()},
+      tracker: 'test',
+    };
     config = {
       domain: 'buckets-o-stuff.myshopify.com',
       storefrontAccessToken: 123,
     };
     client = ShopifyBuy.buildClient(config);
+    client.config.domain = 'test-domain.myshopify.com';
     sinon.stub(client.shop, 'fetchInfo').resolves(shopFixture);
-    ui = new UI(client, {});
     script = document.createElement('script');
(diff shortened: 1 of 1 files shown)


test/unit/ui.js
```diff
@@ -1,124 +1,215 @@
...
-      assert(shopInfo.notCalled);
     });
-  });
 
-  describe('createCart', () => {
+    it('creates an empty array for iframeComponents', () => {
+      assert.deepEqual(ui.iframeComponents, []);
+    });
+
+    it('creates an empty components object holding product, cart, collection, productSet, modal, and toggle', () => {
```
spencercanner: The object isn't really empty here, but just the individual arrays. Maybe something like `creates a components object holding empty arrays for product, cart, collection, productSet, modal, and toggle` would be more explicit

test/unit/ui.js
```diff
@@ -1,124 +1,215 @@
...
+    afterEach(() => {
+      ui = null;
+      resizeSpy.restore();
+      hostClickSpy.restore();
+      escSpy.restore();
+      postMessageSpy.restore();
+      appendStyleTagSpy.restore();
+    });
+
+    it('sets client and domain from passed in client and client.config.domain', () => {
```
spencercanner: We're using `params` to refer to the constructor's parameters in a subsequent test's name (`sets stylesOverrides from params`). Could we use that language here as well for consistency?

test/unit/ui.js
```diff
@@ -1,124 +1,215 @@
...
+        collection: ProductSet,
+        productSet: ProductSet,
+        toggle: CartToggle,
+      };
+      assert.deepEqual(ui.componentTypes, componentTypes);
+    });
 
-    beforeEach(() => {
-      initStub = sinon.stub(Cart.prototype, 'init').resolves();
+    it('sets an error reporter from integrations props', () => {
```
spencercanner: The test below uses the variable name in camel case (`sets stylesOverrides from params`). Here we could be using errorReporter in camel case as well. Also, the word `props` could be replaced with `params` for consistency.

test/unit/ui.js
```diff
@@ -1,124 +1,215 @@
...
+        collection: [],
+        productSet: [],
+        modal: [],
+        toggle: [],
+      };
+      assert.deepEqual(ui.components, emptyComponentsObj);
+    });
 
-    let initStub;
+    it('creates component types object for product, cart, collection, productSet, modal, toggle', () => {
```
spencercanner: Same comment as below with respect to the camel case variable name

test/unit/ui.js
```diff
@@ -1,124 +1,215 @@
...
-    id: 123,
-    options: {}
-  };
 
   beforeEach(() => {
+    integrations = {
+      errorReporter: {notifyException: sinon.spy()},
+      tracker: 'test',
+    };
     config = {
```
spencercanner: Since this is a constant variable, could this be moved outside the beforeEach?

PR #535, opened 2018-11-12
Update ui.js Cart Tests

jameskim08:
- Grouped cart methods together
- Switched from using .then() to async await
- Updated createCart method
- Added tests for toggleCart, closeCart, and openCart

### DIFF
--- a/test/unit/ui.js
+++ b/test/unit/ui.js
@@ -133,36 +133,6 @@ describe('ui class', () => {
       ui = new UI(client, integrations);
     });
 
-    describe('createCart', () => {
-      let initStub;
-
-      beforeEach(() => {
-        initStub = sinon.stub(Cart.prototype, 'init').resolves();
-      });
-
-      afterEach(() => {
-        initStub.restore();
-      });
-
-      describe('when no cart exists', () => {
-        it('creates a new cart', () => {
-          return ui.createCart({options: {}}).then(() => {
-            assert.equal(1, ui.components.cart.length, 'cart array has 1 item');
-            ui.destroyComponent('cart', ui.components.cart[0].model.id);
-          });
-        });
-      });
-
-      describe('when a cart exists', () => {
-        it('does not create a second cart', () => {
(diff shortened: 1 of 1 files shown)


test/unit/ui.js
```diff
@@ -212,5 +182,143 @@ describe('ui class', () => {
...
+        beforeEach(() => {
+          initStub = sinon.stub(Cart.prototype, 'init').resolves();
+        });
+
+        afterEach(() => {
+          initStub.restore();
+        });
+
+        describe('when no cart exists', () => {
+          it('creates a new cart', async () => {
```
melissaluu: I think we've missed testing the return value in this scenario

test/unit/ui.js
```diff
@@ -212,5 +182,143 @@ describe('ui class', () => {
...
+          it('returns the first cart', async () => {
+            const createCartConfig = {toggles: [1, 2, 3]};
+            const response = await ui.createCart(createCartConfig);
+            assert.equal(response, ui.components.cart[0]);
+          });
+        });
+      });
+
+      describe('closeCart', () => {
+        it('calls cart.close for every visible cart and calls restoreFocus after each close', () => {
```
melissaluu: Let's rewrite the descriptions to remove direct references to the functions called and focus the message on the intended behaviour.

test/unit/ui.js
```diff
@@ -212,5 +182,143 @@ describe('ui class', () => {
...
+          initStub.restore();
+        });
+
+        describe('when no cart exists', () => {
+          it('creates a new cart', async () => {
+            await ui.createCart({options: {}});
+            assert.equal(1, ui.components.cart.length);
+            assert.instanceOf(ui.components.cart[0], Cart);
+            assert.calledOnce(initStub);
+            ui.destroyComponent('cart', ui.components.cart[0].model.id);
```
spencercanner: Is there a way we can remove the dependency on `destroyComponent` by stubbing something out?

PR #537, opened 2018-11-12
Updated modal methods, setActiveEl, and restoreFocus in ui.js

jameskim08:
- added tests for modal methods
- removed dependency on destroyComponent
- created tests for setActiveEl and restoreFocus

### DIFF
--- a/test/unit/ui.js
+++ b/test/unit/ui.js
@@ -132,8 +132,20 @@ describe('ui class', () => {
   });
 
   describe('prototype methods', () => {
+    let appendChildStub;
+
     beforeEach(() => {
       ui = new UI(client, integrations);
+      appendChildStub = sinon.stub(document.body, 'appendChild').returns({
+        parentNode: {
+          insertBefore: sinon.spy(),
+          removeChild: sinon.spy(),
+        },
+      });
+    });
+
+    afterEach(() => {
+      appendChildStub.restore();
     });
 
     describe('component methods', () => {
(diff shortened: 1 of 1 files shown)


test/unit/ui.js
```diff
@@ -423,5 +427,123 @@ describe('ui class', () => {
...
+          assert.instanceOf(ui.components.modal[0], Modal);
+        });
+
+        it('does not create a modal if it already exists', () => {
+          ui.components.modal = [{options: {}}];
+          ui.createModal({options: {}});
+          assert.equal(1, ui.components.modal.length);
+        });
+
+        it('returns the newly created modal', () => {
```
melissaluu: As per our IRL convo, let's include these test conditions to the previous tests to maintain context.

test/unit/ui.js
```diff
@@ -423,5 +427,123 @@ describe('ui class', () => {
...
+        ui.restoreFocus();
+        assert.notCalled(focusSpy);
+      });
+
+      it('does not focus on activeEl if cart is open', () => {
+        ui = Object.defineProperty(ui, 'modalOpen', {
+          value: false,
+          writable: false,
+        });
+        ui = Object.defineProperty(ui, 'cartOpen', {
```
spencercanner: Could we do the `Object.defineProperty` inside the `beforeEach` and only set `writable: true` without a value? This would clean up the subsequent tests because we could plainly set the values using `ui.modalOpen = true`

PR #540, opened 2018-11-14
Update Tests for Private Methods in UI.js

jameskim08:
- Created tests for private methods in ui.js
- Changed a spy into a stub
- Cleared modal and carts after creation due to a bug with cleanup
- Changed UI.prototype to ui in stubs to be consistent through the test

### DIFF
--- a/test/unit/ui.js
+++ b/test/unit/ui.js
@@ -7,6 +7,7 @@ import ProductSet from '../../src/components/product-set';
 import Cart from '../../src/components/cart';
 import CartToggle from '../../src/components/toggle';
 import * as browserFeatures from '../../src/utils/detect-features';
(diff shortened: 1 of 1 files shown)


test/unit/ui.js
```diff
@@ -702,5 +720,293 @@ describe('ui class', () => {
...
+            parentNode: {
+              tagName: 'HEAD',
+            },
+          };
+          ui._queryEntryNode();
+          assert.calledOnce(appendToBodyStub);
+          assert.calledWith(appendToBodyStub, div);
+        });
+
+        it('appends to body if the the entry\'s parent node is HTML', () => {
```
spencercanner: two `the`s in the title

PR #543, opened 2018-11-14
Updated constructor tests in component

jameskim08:
- extracted instantiation to a beforeEach
- updated existing constructor tests
- created new constructor tests

### DIFF
--- a/test/unit/component.js
+++ b/test/unit/component.js
@@ -1,19 +1,19 @@
 import Component from '../../src/component';
 import View from '../../src/view';
 import Updater from '../../src/updater';
-import componentDefaults from '../../src/defaults/components';
-
+import * as componentDefaults from '../../src/defaults/components';
+import defaultMoneyFormat from '../../src/defaults/money-format';
 
 describe('Component class', () => {
   describe('constructor', () => {
     let component;
+    let componentDefaultsStub;
     const config = {
       id: 'id',
       handle: 'handle',
       storefrontId: 'sfid',
       debug: 'debug',
       cartNode: 'cartNode',
-      moneyFormat: '${{amount}}',
       modalNode: 'modalNode',
       toggles: 'toggles',
       node: document.createElement('div'),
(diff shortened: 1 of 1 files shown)


test/unit/component.js
```diff
@@ -1,189 +1,211 @@
...
     it('sets globalConfig based on passed in config', () => {
-      let component = new Component({
-        id: 1234
-      });
-      assert.equal(component.globalConfig.moneyFormat, '${{amount}}');
+      const expectedObj = {
+        debug: 'debug',
+        cartNode: 'cartNode',
+        moneyFormat: '${{amount}}',
+        modalNode: 'modalNode',
```
melissaluu: Since `moneyFormat` is calling a function, we should stub it here.

test/unit/component.js
```diff
@@ -1,189 +1,211 @@
...
-      let component = new Component({
-        id: 1234,
-        handle: 'lol',
-        node: node,
-      });
+    let component;
+    const node = document.createElement('div');
+
+    beforeEach(() => {
+      component = new Component({
```
melissaluu: Let's abstract this out to a constant and use the constant reference in our assertions

test/unit/component.js
```diff
@@ -1,189 +1,212 @@
...
 
-  describe('get name()', () => {
-    it('returns name based on handle or id', () => {
-      let component = new Component({id: 1234});
-      assert.equal(component.name, 'frame-undefined-1234', 'uses ID if ID is set');
-      component = new Component({handle: 'lol'});
-      assert.equal(component.name, 'frame-undefined-lol', 'uses handle if handle is set');
+    it('sets config from merging config.options with componentDefaults', () => {
+      assert.equal(component.config.product.buttonDestination, config.options.product.buttonDestination);
+      assert.deepEqual(component.config.cart, componentDefaults.cart);
```
spencercanner: Is it possible to stub out the componentDefaults' default, and do a deepEquals on some mocked return?

PR #544, opened 2018-11-15
Update Tests for Prototype Methods in component.js

jameskim08:
Added tests for getters
Added test for private methods
Updated existing tests to meet standards
Added missing tests for existing tests

### DIFF
--- a/test/unit/component.js
+++ b/test/unit/component.js
@@ -2,6 +2,9 @@ import Component from '../../src/component';
 import View from '../../src/view';
 import Updater from '../../src/updater';
 import * as componentDefaults from '../../src/defaults/components';
+import * as logNotFound from '../../src/utils/log-not-found';
+import * as logger from '../../src/utils/logger';
+import * as isFunction from '../../src/utils/is-function';
 import defaultMoneyFormat from '../../src/defaults/money-format';
 
 describe('Component class', () => {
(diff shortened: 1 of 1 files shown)


test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          try {
+            await component.init(data);
+          } catch (error) {
+            assert.equal(error, setupError);
+          }
+          assert.calledOnce(logNotFoundStub);
+          assert.calledWith(logNotFoundStub, component);
+        });
+
+        it('does not log a not found error if the error message is not "Not Found"', async () => {
```
melissaluu: Technically, it's checking to see if the message contains 'Not found'.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
 
-    describe('get selectors()', () => {
-      it('returns classes formatted as css selectors for each component in manifest', () => {
-        const component = new Component({id: 1234});
-        component.typeKey = 'product';
-        assert.equal(component.selectors.product.product, '.shopify-buy__product');
-        assert.equal(component.selectors.option.option, '.shopify-buy__option-select');
+    describe('setupModel()', () => {
+      it('returns passed data', async () => {
+        const model = await component.setupModel({test: 'test'});
```
melissaluu: Abstracting the return value would be cleaner.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+        assert.deepEqual(model, {test: 'test'});
+      });
+
+      it('fetches data if data was not passed', async () => {
+        component.fetchData = sinon.stub().resolves({test: 'fetchData'});
```
melissaluu: Same as above, an abstraction would be better in this case.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
-            option: {
-              googleFonts: ['bar'],
-            },
-          },
-        };
-        const component = new Component(config);
-        component.typeKey = 'product';
-        assert.deepEqual(component.googleFonts, ['lol', 'bar']);
+    describe('updateConfig()', () => {
+      it('delegates to updater', () => {
```
melissaluu: Do we want to be more explicit here and explain what the behaviour is?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
 
-    describe('get viewData()', () => {
-      it('returns model and some other properties', () => {
-        const component = new Component({id: 1234});
-        component.typeKey = 'product';
-        component.model = {test: 'lol'};
-        assert.equal(component.viewData.test, 'lol');
-        assert.equal(component.viewData.classes.product.product, 'shopify-buy__product');
+    describe('destroy()', () => {
+      it('delegates to view', () => {
```
melissaluu: Same as above - a more explicit description will help.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
     });
 
-    describe('init()', () => {
-      it('assigns model and initializes view', () => {
-        const component = new Component({id: 1234}, {
-          browserFeatures: {},
+    describe('getters', () => {
+      describe('name', () => {
+        it('returns name based on handle or id', () => {
+          assert.equal(component.name, 'frame-undefined-1234');
```
melissaluu: Let's split these to 2 separate tests for clarity.

test/unit/component.js
spencercanner: Could we put the test object into a variable, and reference it here and in the assert? Same goes for the next test

test/unit/component.js
spencercanner: Could we use this method of defining stubs (`const fetchDataStub = sinon.stub(component, 'fetchData').resolves(...);`) to be consistent with other tests?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          });
+          component.typeKey = 'product';
+          assert.deepEqual(component.styles, {product: {button: {color: 'red'}}});
+        });
+      });
+
+      describe('classes', () => {
+        it('returns classes for each component in manifest', () => {
+          component.typeKey = 'product';
+          assert.equal(component.classes.product.product, 'shopify-buy__product');
```
melissaluu: Where is this value coming from? If possible, we should assert to known values.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
       });
     });
 
-    describe('init()', () => {
-      it('assigns model and initializes view', () => {
-        const component = new Component({id: 1234}, {
-          browserFeatures: {},
+    describe('getters', () => {
+      describe('name', () => {
+        it('returns name based on handle or id', () => {
```
spencercanner: Since we're creating the initial component in the prototype's beforeEach, it is a bit hard to trace where the `1234` comes from. Could we put the id into it's own variable, and use that when we create the component in the beforeEach, and also reference it here for clarity?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          component.typeKey = 'product';
+          assert.deepEqual(component.styles, {product: {button: {color: 'red'}}});
+        });
+      });
+
+      describe('classes', () => {
+        it('returns classes for each component in manifest', () => {
+          component.typeKey = 'product';
+          assert.equal(component.classes.product.product, 'shopify-buy__product');
+          assert.equal(component.classes.option.option, 'shopify-buy__option-select');
```
melissaluu: Is there only 2 types of classes to test?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+              product: {
+                styles: {
+                  button: {
+                    color: 'red',
+                  },
+                },
+              },
+            },
+          });
+          component.typeKey = 'product';
```
melissaluu: Since `typeKey` isn't set directly in the component class (it seems to rely on a inherited class to set it), we should probably set this up at the top so it's clearer.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          assert.equal(component.classes.product.product, 'shopify-buy__product');
+          assert.equal(component.classes.option.option, 'shopify-buy__option-select');
+        });
+      });
+
+      describe('selectors', () => {
+        it('returns classes formatted as css selectors for each component in manifest', () => {
+          component.typeKey = 'product';
+          assert.equal(component.selectors.product.product, '.shopify-buy__product');
```
melissaluu: Same as above regarding to where these values are coming from and how does the manifest affects these selections.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          assert.equal(component.selectors.option.option, '.shopify-buy__option-select');
+        });
+      });
+
+      describe('googleFonts', () => {
+        it('returns google fonts for each component in manifest', () => {
```
melissaluu: Same as above, how does manifest affect this?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+            model: 'model',
+            classes: 'classes',
+          };
+          assert.deepEqual(component.viewData, expectedObj);
+        });
+      });
+
+      describe('morphCallbacks', () => {
+        it('returns an object with the function onBeforeElUpdated', () => {
+          assert.instanceOf(component.morphCallbacks.onBeforeElUpdated, Function);
```
melissaluu: Although it's implied, explicitly testing that the getter returns an object with a property that is a function will be better.

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+        });
+
+        it('calls event if the method passed is a function in the event', () => {
+          const eventSpy = sinon.spy();
+          component = Object.defineProperty(component, 'events', {
+            writable: false,
+            value: {test: eventSpy},
+          });
+
+          const isFunctionStub = sinon.stub(isFunction, 'default').returns(true);
```
melissaluu: Is mocking `isFunction` is necessary? Is eventSpy not a function?

test/unit/component.js
```diff
@@ -82,142 +85,344 @@ describe('Component class', () => {
...
+          assert.deepEqual(component.googleFonts, ['product', 'option']);
+        });
+      });
+
+      describe('viewData', () => {
+        it('returns merged object of model, viewData, classes, and text', () => {
+          component.typeKey = 'product';
+          component.model = {model: 'model'};
+          component = Object.defineProperty(component, 'options', {
+            writable: false,
```
spencercanner: Do we need to explicitly provide a value for `writable`?

test/unit/component.js
jameskim08: isFunction is an import from the utils folder so I thought it would be safer to just not depend on it

test/unit/component.js
jameskim08: I tried that at first, but fetchData does not exist within component.js itself but its children. I was debating if it was worth to create a specific component that holds the fetchData function and stubbing it or approaching it like how it currently is.

test/unit/component.js
```diff
@@ -82,142 +85,374 @@ describe('Component class', () => {
...
-        component.view.destroy = sinon.spy();
-        component.destroy();
-        assert.calledOnce(component.view.destroy);
+    describe('"private" methods', () => {
+      describe('_userEvent()', () => {
+        it('logs to logger if debug is set to true', () => {
+          const infoSpy = sinon.spy();
+          const loggerStub = sinon.stub(logger, 'default').value({info: infoSpy});
+          component.globalConfig.debug = true;
+          component.typeKey = 'key';
```
spencercanner: Do we need to change the typekey here, or could we just use the once set on the prototype?

jameskim08: I added it because the text uses the typeKey so I wanted to be more explicit about it

PR #546, opened 2018-11-15
Update constructor and init in view.js tests

jameskim08:
- abstracted instantiation of component and view to a beforeEach() for prototype methods
- updated constructor tests
- updated init tests

### DIFF
--- a/test/unit/view.js
+++ b/test/unit/view.js
@@ -5,67 +5,110 @@ import Iframe from '../../src/iframe';
 
 describe('View class', () => {
   describe('constructor', () => {
-    const component = new Component({id: 1234});
-    const view = new View(component);
+    let component;
+    let view;
+
+    beforeEach(() => {
+      component = new Component({
+        id: 1234,
+        node: document.createElement('div'),
+      });
+      view = new View(component);
+    });
 
     it('stores component to instance', () => {
       assert.equal(view.component, component);
     });
 
+    it('sets iframe to null', () => {
+      assert.equal(view.iframe, null);
+    });
+
+    it('sets node to component\'s node', () => {
+      assert.equal(view.node, component.node);
+    });
+
     it('creates a template instance', () => {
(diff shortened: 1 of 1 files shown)


test/unit/view.js
```diff
@@ -5,67 +5,105 @@ import Iframe from '../../src/iframe';
...
+      });
 
-      beforeEach(() => {
-        component = new Component({
-          id: 1234,
-          node: document.createElement('div'),
-        }, {browserFeatures: {}});
-        view = new View(component);
-        component.typeKey = 'product';
+      it('create, and loads an returns an Iframe', async () => {
```
spencercanner: Were not testing the return in this test, so maybe say something like `creates and loads an Iframe`.

test/unit/view.js
```diff
@@ -5,67 +5,105 @@ import Iframe from '../../src/iframe';
...
+        assert.calledOnce(addClassStub);
+        assert.calledWith(addClassStub, view.className);
+      });
+
+      it('adds class name dependent on typeKey to component node', async () => {
+        component.typeKey = 'typeKey';
+        await view.init();
+        assert.equal(component.node.className, ' shopify-buy-frame shopify-buy-frame--typeKey');
+      });
+    });
```
spencercanner: I think we're missing a test for the `this.iframe.load()` return

test/unit/view.js
```diff
@@ -5,67 +5,105 @@ import Iframe from '../../src/iframe';
...
+      assert.equal(view.eventsBound, false);
+    });
   });
 
   describe('prototype methods', () => {
+    let component;
+    let view;
+
+    beforeEach(() => {
+      Component.prototype.typeKey = 'product';
```
melissaluu: Since we aren't re-instantiating the component in the subsequent tests, let's set the `typeKey` in the instance of the component instead of modifying the prototype.

PR #548, opened 2018-11-16
Update prototype method tests in view.js part 1

jameskim08:
The following methods in view.js have been updated/added:
- render
- delegateEvents
- append
- destroy
- renderChild
- updateNode
- wrapTemplate

### DIFF
--- a/test/unit/view.js
+++ b/test/unit/view.js
@@ -109,68 +109,140 @@ describe('View class', () => {
     });
 
     describe('render()', () => {
-      it('creates div and updates html', () => {
-        const div = document.createElement('div');
-        const tmplRender = sinon.stub(view.template, 'render').returns('<div>LOL</div>');
-        view._createWrapper = sinon.stub().returns(div);
-        view.resize = sinon.spy();
-        view.updateNode = sinon.spy();
+      let userEventStub;
+      let templateRenderStub;
+      let wrapTemplateStub;
+      let createWrapperStub;
+      let updateNodeStub;
+      let resizeStub;
+      let div;
+      let htmlTemplate;
+
+      beforeEach(() => {
+        div = document.createElement('div');
+        htmlTemplate = '<div>test</div>';
+        userEventStub = sinon.stub(component, '_userEvent');
+        templateRenderStub = sinon.stub(view.template, 'render').returns(htmlTemplate);
+        wrapTemplateStub = sinon.stub(view, 'wrapTemplate');
+        createWrapperStub = sinon.stub(view, '_createWrapper').returns(div);
+        updateNodeStub = sinon.stub(view, 'updateNode');
+        resizeStub = sinon.stub(view, 'resize');
+      });
+
+      afterEach(() => {
+        userEventStub.restore();
+        templateRenderStub.restore();
+        wrapTemplateStub.restore();
+        createWrapperStub.restore();
+        updateNodeStub.restore();
+        resizeStub.restore();
+      });
+
+      it('renders template by passing in viewData and callback function to template\'s render', () => {
         view.render();
-        assert.calledOnce(view._createWrapper);
-        assert.calledOnce(view.resize);
-        assert.calledWith(view.updateNode, div, '<div>LOL</div>');
-        tmplRender.restore();
+        assert.calledOnce(templateRenderStub);
+        assert.deepEqual(templateRenderStub.getCall(0).args[0], {data: component.viewData});
+        templateRenderStub.getCall(0).args[1]('test');
+        assert.calledOnce(wrapTemplateStub);
+        assert.calledWith(wrapTemplateStub, 'test');
+      });
+
+      it('calls component\'s user event befor and after render', () => {
+        view.render();
+        assert.calledTwice(userEventStub);
+        assert.calledWith(userEventStub.getCall(0), 'beforeRender');
+        assert.calledWith(userEventStub.getCall(1), 'afterRender');
+        assert.calledOnce(templateRenderStub);
+      });
+
+      it('creates a wrapper if one does not already exist', () => {
+        view.wrapper = null;
+        view.render();
+        assert.calledOnce(createWrapperStub);
+        assert.equal(view.wrapper, div);
+      });
+
+      it('does not create a new wrapper if one already exists', () => {
+        view.wrapper = 'test';
+        view.render();
+        assert.notCalled(createWrapperStub);
+      });
+
+      it('updates node with wrapper and html then resizes', () => {
+        view.render();
+        assert.calledOnce(updateNodeStub);
+        assert.calledWith(updateNodeStub, view.wrapper, htmlTemplate);
+        assert.calledOnce(resizeStub);
       });
     });
 
     describe('delegateEvents()', () => {
-      beforeEach(() => {
-        component = new Component({
-          id: 1234,
-          node: document.createElement('div'),
-          options: {
-            product: {
-              DOMEvents: {
-                'click .btn'() { return true; },
-                'click'() { return true; },
-              },
-            },
-          },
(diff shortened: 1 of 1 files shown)


test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
-        assert.calledWith(view.updateNode, div, '<div>LOL</div>');
-        tmplRender.restore();
+        assert.calledOnce(templateRenderStub);
+        assert.deepEqual(templateRenderStub.getCall(0).args[0], {data: component.viewData});
+        templateRenderStub.getCall(0).args[1]('test');
+        assert.calledOnce(wrapTemplateStub);
+        assert.calledWith(wrapTemplateStub, 'test');
+      });
+
+      it('calls component\'s user event befor and after render', () => {
```
spencercanner: Typo: `before`

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
+        assert.calledOnce(closeComponentsOnEscStub);
+      });
+
+      it('calls _on for each DOM event with a selector', () => {
         view.delegateEvents();
-        assert.calledWith(onStub, 'click', '.btn', sinon.match.func);
-        assert.calledWith(view.wrapper.addEventListener, 'click', sinon.match.func);
+        assert.calledOnce(onStub);
+        assert.calledWith(onStub, 'click', '.btn');
+        onStub.getCall(0).args[2]();
```
spencercanner: Can we call the function with two arguments?

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
+        assert.calledOnce(component.DOMEvents['click .btn']);
+      });
+
+      it('adds event listener for DOM event without a selector', () => {
+        view.delegateEvents();
+        assert.calledWith(view.wrapper.addEventListener, 'click');
+        view.wrapper.addEventListener.getCall(0).args[1]();
```
spencercanner: Same as above, can we pass in an `evt`, and check calledWith for the DOMEvents click?

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
+        assert.calledOnce(component.DOMEvents.click);
       });
 
       it('bind events if eventsBound is false', () => {
-        view.wrapper = {
-          addEventListener: sinon.spy(),
-        };
         view.delegateEvents();
         assert.called(view.wrapper.addEventListener);
```
spencercanner: Could we check the `_on` and `closeComponentsOnEsc` stubs as well, since those functions are also supposed to do event binding?

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
+      });
+
+      it('adds event listener for DOM event without a selector', () => {
+        view.delegateEvents();
+        assert.calledWith(view.wrapper.addEventListener, 'click');
+        view.wrapper.addEventListener.getCall(0).args[1]();
+        assert.calledOnce(component.DOMEvents.click);
       });
 
       it('bind events if eventsBound is false', () => {
```
spencercanner: Typo: `binds`

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
       });
 
-      it('prevents rebinding if events already bound', () => {
-        view.wrapper = {
-          addEventListener: sinon.spy(),
-        };
+      it('prevents rebinding if events are already bound', () => {
         view.eventsBound = true;
         view.delegateEvents();
         assert.notCalled(view.wrapper.addEventListener);
```
spencercanner: Same as above, can we check the `_on` and `closeComponentsOnEsc` stubs as well?

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
+        assert.calledOnce(component.DOMEvents.click);
       });
 
       it('bind events if eventsBound is false', () => {
-        view.wrapper = {
-          addEventListener: sinon.spy(),
-        };
         view.delegateEvents();
         assert.called(view.wrapper.addEventListener);
       });
```
spencercanner: Can we also check that it sets `eventsBound` to true

test/unit/view.js
```diff
@@ -5,143 +5,239 @@ import Iframe from '../../src/iframe';
...
 
-      beforeEach(() => {
-        component = new Component({
-          id: 1234,
-          node: document.createElement('div'),
-        }, {browserFeatures: {}});
-        view = new View(component);
+        view.delegateEvents();
+        view.iframe.el.onload();
+        assert.calledOnce(reloadIframeStub);
```
spencercanner: Can we check that onload is set to null?

test/unit/view.js
```diff
@@ -151,101 +247,85 @@ describe('View class', () => {
           },
         };
         view.append(div);
+        assert.calledOnce(view.iframe.document.body.appendChild);
```
spencercanner: For consistency with other tests, can we put the spy into it's own const and assert with that instead?

test/unit/view.js
```diff
@@ -151,101 +247,85 @@ describe('View class', () => {
...
-        }, {browserFeatures: {}});
-        const view = new View(component);
         view.node = {
           parentNode: {
             removeChild: sinon.spy(),
           },
         };
         view.destroy();
         assert.calledOnce(view.node.parentNode.removeChild);
+        assert.calledWith(view.node.parentNode.removeChild, view.node);
```
spencercanner: Same comment as above, can we move out the spy into it's own const, and use that in the assert?

test/unit/view.js
```diff
@@ -151,101 +247,85 @@ describe('View class', () => {
...
         const template = {
-          render: sinon.stub().returns('<h1>BUY MY BUTTONS lol</h1>')
+          render: sinon.stub().returns(html),
         };
-        view.updateNode = sinon.spy();
-        view.wrapper = document.createElement('div');
-        const childNode = document.createElement('div');
-        childNode.className = 'title';
-        view.wrapper.appendChild(childNode);
+        view.wrapper = {querySelector: sinon.stub().returns(node)};
```
spencercanner: Can we format this like `template` above, and also move out the stubs into their own variables for consistency?

test/unit/view.js
```diff
@@ -151,101 +247,85 @@ describe('View class', () => {
...
-      const view = new View(component);
-
       it('updates contents of node', () => {
         const div = document.createElement('div');
-        div.innerHTML = '<h1>OLD TEXT</h1>';
-        const html = '<h1>SO FRESH</h1>';
+        div.innerHTML = '<h1>old</h1>';
+        const html = '<h1>new</h1>';
         view.updateNode(div, `<div>${html}</div>`);
         assert.equal(div.innerHTML, html);
```
spencercanner: This test is a bit inconsistent with others, since we're testing that morphdom works correctly here, not that we're calling it with the expected things.

test/unit/view.js
```diff
@@ -194,61 +203,74 @@ describe('View class', () => {
...
-        view.wrapper.addEventListener.getCall(0).args[1]();
-        assert.calledOnce(component.DOMEvents.click);
+        assert.calledWith(addEventListenerSpy, 'click');
+        addEventListenerSpy.getCall(0).args[1](event);
+        assert.calledOnce(clickSpy);
+        assert.calledWith(clickSpy, event);
       });
 
-      it('bind events if eventsBound is false', () => {
+      it('binds events and sets eventsBoudn to true if eventsBound is false', () => {
```
spencercanner: Typo with `eventsBoudn`

PR #549, opened 2018-11-19
Update Prototype Method Tests in view.js Part 2

jameskim08:
The following tests have been created/updated:
- reloadIframe
- addClass
- removeClass
- setFocus
- animateRemoveNode
- removeNode

### DIFF
--- a/test/unit/view.js
+++ b/test/unit/view.js
@@ -2,6 +2,7 @@ import View from '../../src/view';
 import Component from '../../src/component';
 import Template from '../../src/template';
 import Iframe from '../../src/iframe';
+import * as elementClass from '../../src/utils/element-class';
 
 describe('View class', () => {
   describe('constructor', () => {
@@ -262,6 +263,26 @@ describe('View class', () => {
       });
     });
 
+    describe('reloadIframe()', () => {
+      it('removes iframe and initializes component', () => {
+        const el = document.createElement('div');
+        view.iframe = {el};
+        view.wrapper = {};
+        const removeChildStub = sinon.stub(view.node, 'removeChild');
+        const initStub = sinon.stub(view.component, 'init');
+
+        view.reloadIframe();
+        assert.calledOnce(removeChildStub);
+        assert.calledWith(removeChildStub, el);
+        assert.isNull(view.wrapper);
+        assert.isNull(view.iframe);
+        assert.calledOnce(initStub);
+
+        removeChildStub.restore();
+        initStub.restore();
+      });
+    });
+
     describe('append()', () => {
       it('appends to document if iframe exists', () => {
         const appendChildSpy = sinon.spy();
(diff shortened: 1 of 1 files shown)


test/unit/view.js
```diff
@@ -289,6 +310,64 @@ describe('View class', () => {
...
+        assert.calledOnce(addClassSpy);
+        assert.calledWith(addClassSpy, 'test-class');
+        assert.notCalled(addClassToElementStub);
+      });
+
+      it('adds class to element if iframe does not exist', () => {
+        view.iframe = null;
+        view.addClass('test-class');
+        assert.calledOnce(addClassToElementStub);
+        assert.calledWith(addClassToElementStub, 'test-class');
```
melissaluu: I think `addClassToElement` takes multiple parameters.

test/unit/view.js
```diff
@@ -289,6 +310,64 @@ describe('View class', () => {
...
+        assert.calledOnce(removeClassSpy);
+        assert.calledWith(removeClassSpy, 'test-class');
+        assert.notCalled(removeClassFromElementStub);
+      });
+
+      it('removes class from element if iframe does not exist', () => {
+        view.iframe = null;
+        view.removeClass('test-class');
+        assert.calledOnce(removeClassFromElementStub);
+        assert.calledWith(removeClassFromElementStub, 'test-class');
```
melissaluu: Similar to the above, I think `removeClassFromElement` takes multiple parameters.

test/unit/view.js
```diff
@@ -354,116 +433,172 @@ describe('View class', () => {
...
+        assert.calledWith(addClassToElementStub, 'is-hidden', document.getElementById(123));
+      });
+
+      it('removes node on animationend event if animation is supported and element has a parent node', () => {
+        view.component.props.browserFeatures.animation = true;
+        view.animateRemoveNode(123);
+        node.dispatchEvent(event);
+        assert.calledOnce(removeNodeStub);
+        assert.calledWith(removeNodeStub, document.getElementById(123));
       });
```
melissaluu: I think we can add a test to cover the case of when an element doesn't have a parent node.

test/unit/view.js
```diff
@@ -354,116 +433,172 @@ describe('View class', () => {
...
+        removeNodeStub.restore();
+        addClassToElementStub.restore();
       });
 
-      it('adds event listener on animationend if browser supports it', () => {
+      it('adds is-hidden class to element', () => {
         view.animateRemoveNode(123);
-        assert.calledWith(node.addEventListener, 'animationend');
+        assert.calledOnce(addClassToElementStub);
+        assert.calledWith(addClassToElementStub, 'is-hidden', document.getElementById(123));
```
spencercanner: Does using `node` work here, or do we need to query the DOM?

test/unit/view.js
jameskim08: Currently, I can't figure out how to get it into the if block. This is because it grabs the node from the document and then checks if the node has a parentNode. If the node is from the document, then the document itself would be the parentNode.

PR #550, opened 2018-11-19
Update Getters and private methods tests in view.js

jameskim08:
Updated tests for getters and private methods in view.js

### DIFF
--- a/test/unit/view.js
+++ b/test/unit/view.js
@@ -601,6 +601,238 @@ describe('View class', () => {
         renderStub.restore();
       });
     });
+
+    describe('getters', () => {
+      describe('outerHeight', () => {
+        let getPropertyValueSpy;
+
+        beforeEach(() => {
+          view.wrapper = document.createElement('div');
+          getPropertyValueSpy = sinon.spy(CSSStyleDeclaration.prototype, 'getPropertyValue');
+        });
+
+        afterEach(() => {
+          getPropertyValueSpy.restore();
+        });
+
+        it('returns wrapper height if there is no styling on wrapper', () => {
+          view.wrapper = {clientHeight: 10};
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(null);
+          assert.equal(view.outerHeight, '10px');
+          getComputedStyleStub.restore();
+        });
+
+        it('returns the height of the wrapper set in style', () => {
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
+          view.wrapper.style.height = '50px';
+          assert.equal(view.outerHeight, '50px');
+          getComputedStyleStub.restore();
+        });
+
+        it('returns wrapper\'s client height if height is not set in style', () => {
+          view.wrapper = {
+            clientHeight: 20,
+            style: {
+              height: '',
+              getPropertyValue: sinon.spy(),
+            },
+          };
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
+          assert.equal(view.outerHeight, '20px');
+          getComputedStyleStub.restore();
+        });
+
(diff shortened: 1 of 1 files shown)


test/unit/view.js
```diff
@@ -601,6 +601,186 @@ describe('View class', () => {
...
+
+        it('returns the height of the wrapper set in style', () => {
+          view.wrapper = document.createElement('div');
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
+          view.wrapper.style.height = '50px';
+          assert.equal(view.outerHeight, '50px');
+          getComputedStyleStub.restore();
+        });
+
+        it('returns wrapper\'s client height if height is not set in style', () => {
```
melissaluu: I think we're missing some coverage regarding this condition.

test/unit/view.js
```diff
@@ -601,6 +601,186 @@ describe('View class', () => {
...
+          assert.calledOnce(appendStub);
+          assert.calledWith(appendStub, mockWrapper);
+        });
+
+        it('returns the newly created wrapper', () => {
+          assert.equal(view._createWrapper(), mockWrapper);
+        });
+
+        it('adds component type class to wrapper', () => {
+          assert.equal(view._createWrapper().className, 'shopify-buy__product');
```
melissaluu: Can we identify where this class name is coming from? It seems a bit arbitrary...

test/unit/view.js
```diff
@@ -601,6 +601,242 @@ describe('View class', () => {
...
+          assert.calledOnce(appendStub);
+          assert.calledWith(appendStub, mockWrapper);
+        });
+
+        it('returns the newly created wrapper', () => {
+          assert.equal(view._createWrapper(), mockWrapper);
+        });
+
+        it('adds component type class to wrapper', () => {
+          component.typeKey = 'typeKey';
```
melissaluu: Can we use a more "real" typeKey value? Using `typeKey` here is a bit confusing...

test/unit/view.js
```diff
@@ -601,6 +601,232 @@ describe('View class', () => {
...
+
+        it('calls getPropertyValue() twice if height is set to auto in style', () => {
+          view.wrapper.style.height = 'auto';
+          assert.equal(view.outerHeight, '0px');
+          assert.calledTwice(getPropertyValueSpy);
+          assert.calledWith(getPropertyValueSpy, 'height');
+        });
+
+        it('calls getPropertyValue() once if height is set to a value in style', () => {
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
```
spencercanner: Why we need this stub here, but not in the few tests above?

jameskim08: The stub is used to grab the view.wrapper.style.height in the line below because the original function does not read it properly

spencercanner: If we were to stub out `getComputedStyle` in the tests above, would be able to use `assert.equal(...)` instead of `assert.isOk`?

jameskim08: Yep!

test/unit/view.js
```diff
@@ -601,6 +601,238 @@ describe('View class', () => {
...
+          };
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
+          assert.equal(view.outerHeight, '20px');
+          getComputedStyleStub.restore();
+        });
+
+        it('calls getPropertyValue() twice if height is not set in style', () => {
+          const getComputedStyleStub = sinon.stub(window, 'getComputedStyle').returns(view.wrapper.style);
+          view.wrapper.style.height = '';
+          assert.isOk(view.outerHeight, `${view.wrapper.clientHeight}px`);
```
spencercanner: Can we switch the `isOk` back to `equal` in these tests?

PR #553, opened 2018-11-20
Update template.js tests

jameskim08:
- Refactored format to be consistent with other tests
- Tested behaviour of Mustache in render() due to being unable to test external libraries
- Renamed variable names and values
- Created test for constructor
- Edited test descriptions

### DIFF
--- a/test/unit/template.js
+++ b/test/unit/template.js
@@ -1,60 +1,63 @@
 import Template from '../../src/template';
 
-const contents = {
-  title: true,
-  button: true,
-  description: false,
-}
-const templates = {
-  button: '<button>BUTTON</button>',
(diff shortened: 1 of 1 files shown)


test/unit/template.js
```diff
@@ -1,60 +1,68 @@
...
-    describe('without callback', () => {
-      it('it puts data into the strings', () => {
-        const expectedString = '<h1>BUY MY BUTTONS fool</h1><button>BUTTON</button>';
-        const data = {
-          name: 'fool'
-        }
-        const output = template.render({data: data});
-        assert.equal(expectedString, output);
+      beforeEach(() => {
+        expectedString = '<h1>title test</h1><button>button</button>';
```
spencercanner: Do these need to be in a `beforeEach` if they're just constant values?

PR #554, opened 2018-11-21
Update iframe.js tests part 1

jameskim08:
- Reformatted test to match format of other tests (constructor and prototype methods separated)
- Created tests for constructor
- Refactored and created tests for load()
- Refactored and created tests for loadFonts()
- assert.isUndefined is used to check for return resolve() calls
- commented out loadFontScript() tests (will be refactored in future PR)

### DIFF
--- a/test/unit/iframe.js
+++ b/test/unit/iframe.js
@@ -1,94 +1,268 @@
 import Iframe from '../../src/iframe';
 
 const defaultCSS = '* { box-sizing: border-box; }';
-const customCSS = '@media (max-width: 100px) { .product { background: blue; }  } .product:hover .btn { background: orange; } .btn:hover { color: green; } .btn { color: red; }';
-
-let iframe;
-let parent;
+// const customCSS = '@media (max-width: 100px) { .product { background: blue; }  } .product:hover .btn { background: orange; } .btn:hover { color: green; } .btn { color: red; }';
+const configObject = {
+  classes: {
+    product: {
+      button: 'btn',
+      product: 'product',
+    },
+  },
+  customStyles: {
+    product: {
+      product: {
+        '@media (max-width: 100px)': {
+          background: 'blue',
+        },
+        ':hover': {
+          button: {
+            background: 'orange',
+          },
+        },
+      },
+      button: {
+        color: 'red',
+        ':hover': {
+          color: 'green',
+        },
+      },
+    },
+  },
+  stylesheet: defaultCSS,
+  browserFeatures: {
+    transition: true,
+    animation: true,
+    transform: true,
+  },
+  name: 'frame-product',
+};
 
 describe('Iframe class', () => {
+  let parent;
+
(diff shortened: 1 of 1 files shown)


test/unit/iframe.js
```diff
@@ -1,94 +1,271 @@
...
-      return iframe.load().then(() => {
-        assert.include(iframe.css, defaultCSS, 'css is formatted correctly');
-        assert.include(iframe.css, customCSS, 'appends custom css');
+    describe('load()', () => {
+      let loadFontsStub;
+      let appendStyleTagStub;
+
+      beforeEach(() => {
+        loadFontsStub = sinon.stub(iframe, 'loadFonts').resolves();
+        appendStyleTagStub = sinon.stub(iframe, 'appendStyleTag').resolves();
```
spencercanner: Does this need to resolve?

test/unit/iframe.js
```diff
@@ -1,94 +1,271 @@
...
+    let parent;
+    let iframe;
+
     beforeEach(() => {
-      scriptTags = document.head.getElementsByTagName('script').length;
-      iframe.googleFonts = ['Lato'];
-      return iframe.load();
+      parent = document.createElement('div');
+      parent.setAttribute('id', 'fixture');
+      document.body.appendChild(parent);
```
spencercanner: Could these be moved into a global `beforeEach`, since they're in the repeated in the constructor and prototype methods describe blocks?

test/unit/iframe.js
```diff
@@ -1,94 +1,271 @@
...
+        assert.isUndefined(response);
+      });
+
+      it('loads WebFont with google fonts, a function on font render, and content window', async () => {
+        iframe.el = {contentWindow: {}};
+        await iframe.loadFonts();
+        assert.deepEqual(loadSpy.getCall(0).args[0].google, {families: iframe.googleFonts});
+        assert.instanceOf(loadSpy.getCall(0).args[0].fontactive, Function);
+        assert.isUndefined(loadSpy.getCall(0).args[0].fontactive());
+        assert.equal(loadSpy.getCall(0).args[0].context, iframe.el.contentWindow);
```
spencercanner: Can we extract the `loadSpy.getCall(0).args[0]` into a variable?

test/unit/iframe.js
```diff
@@ -1,94 +1,271 @@
...
+        await iframe.load();
+        assert.equal(parent.children.length, 1);
+        assert.equal(parent.children[0].tagName, 'IFRAME');
+      });
+
+      it('loads fonts then append style tag on iframe load', async () => {
+        await iframe.load();
+        assert.calledOnce(loadFontsStub);
+        assert.calledOnce(appendStyleTagStub);
+        await iframe.el.onload();
```
melissaluu: Since `appendChild()` will trigger the element's `onload()`, I think we can omit the explicit `iframe.el.onload()`.

PR #557, opened 2018-11-22
Update iframe.js getter tests

jameskim08:
Added tests for:
- get width
- get document
- get conditionalCSS
- get css

get customStyles due to how complicated it is will be saved for another time. However, get css currently indirectly tests customStyles.

### DIFF
--- a/test/unit/iframe.js
+++ b/test/unit/iframe.js
@@ -1,94 +1,268 @@
 import Iframe from '../../src/iframe';
 
 const defaultCSS = '* { box-sizing: border-box; }';
-const customCSS = '@media (max-width: 100px) { .product { background: blue; }  } .product:hover .btn { background: orange; } .btn:hover { color: green; } .btn { color: red; }';
-
-let iframe;
-let parent;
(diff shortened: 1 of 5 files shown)


test/unit/iframe.js
```diff
@@ -1,94 +1,500 @@
...
+        it('returns conditionalStyles if browser features does not include animation', () => {
+          iframe.browserFeatures.animation = false;
+          assert.equal(iframe.conditionalCSS, conditionalStyles);
+        });
+      });
+
+      describe('css', () => {
+        it('returns properly formatted CSS', () => {
+          const customCSS = '@media (max-width: 100px) { .product { background: blue; }  } .product:hover .btn { background: orange; } .btn:hover { color: green; } .btn { color: red; }';
+          assert.include(iframe.css, iframe.stylesheet);
```
melissaluu: We should set `iframe.css` to a `const` so we don't need to call it multiple times

jameskim08:
Closed this because it is merging to the wrong branch.

PR #563, opened 2018-11-29
Refactor existing product component tests part 1

jameskim08:
- Updated/created tests for constructor, init, sdkFetch, and updateVariant
- moved sdkFetch out of other prototype methods because it uses its own Product instantiation
- removed a constructor test that depended on normalizeConfig

### DIFF
--- a/test/unit/product/product-component.js
+++ b/test/unit/product/product-component.js
@@ -32,7 +32,7 @@ describe('Product Component class', () => {
 
     beforeEach(() => {
       normalizeConfigStub = sinon.stub(normalizeConfig, 'default').callsFake((oldConfig) => {
-        oldConfig.normalized = true;
+        oldConfig.storefrontId = 'normalizedId';
         return oldConfig;
       });
       product = new Product(constructorConfig, props);
@@ -40,13 +40,12 @@ describe('Product Component class', () => {
 
     afterEach(() => {
       normalizeConfigStub.restore();
-      constructorConfig.normalized = null;
     });
 
     it('normalizes config', () => {
       assert.calledOnce(normalizeConfigStub);
       assert.calledWith(normalizeConfigStub, constructorConfig);
-      assert.isTrue(constructorConfig.normalized);
+      assert.equal(constructorConfig.storefrontId, 'normalizedId');
     });
 
     it('sets typeKey to product', () => {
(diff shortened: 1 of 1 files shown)


test/unit/product/product-component.js
```diff
@@ -3,715 +3,926 @@ import Checkout from '../../../src/components/checkout';
...
+        let createCartStub;
+        let renderStub;
 
-        product.onButtonClick(evt, target);
-        assert.calledOnce(addToCartStub);
-        assert.calledWith(addToCartStub, product.selectedVariant, 1111);
-        addToCartStub.restore();
-      });
+        beforeEach(() => {
+          createCartStub = sinon.stub(product.props, 'createCart').resolves('test');
```
melissaluu: Can we change the value of the resolved value to something that resembles a cart object? The use of `test` is a bit confusing since `init` is also called with `test` as the `data` param.

test/unit/product/product-component.js
```diff
@@ -3,715 +3,926 @@ import Checkout from '../../../src/components/checkout';
...
+            await product.init('test');
           });
-        });
 
-        const checkout = new Checkout(product.config);
-        const evt = new Event('click shopify-buy__btn--parent');
-        const target = 'shopify-buy__btn--parent';
+          afterEach(() => {
+            superInitStub.restore();
+          });
```
melissaluu: Are we missing a test around the returned value from `init`?

test/unit/product/product-component.js
```diff
@@ -3,715 +3,926 @@ import Checkout from '../../../src/components/checkout';
...
+          userEventStub = sinon.stub(product, '_userEvent');
+        });
 
-    describe('openModal()', () => {
-      describe('if modal exists', () => {
-        let modalInitSpy;
+        afterEach(() => {
+          renderStub.restore();
+          userEventStub.restore();
+        });
```
melissaluu: Do we need to test the case when the param `option` doesn't actually exists?

test/unit/product/product-component.js
```diff
@@ -3,715 +3,926 @@ import Checkout from '../../../src/components/checkout';
...
+            product.updateVariant('Size', 'large');
+            assert.equal(product.selectedImage, product.model.images[0]);
           });
         });
 
-        describe('if image and cached image are different', () => {
-          beforeEach(async () => {
-            product.config.product.width = '100px';
-            await product.init(testProductCopy);
+        it('sets selected image to the cached image in model if variant does not exist', () => {
```
melissaluu: I think this test needs to be outside of the above `if variant exists` `describe` block?

jameskim08: the test seems to already exist outside of the 'if variant exists' block

PR #570, opened 2018-12-11
Update product.js getters tests Part 1

jameskim08:
Updated existing tests for the following getters:
- shouldUpdateImage
- currentImage
- viewData
- buttonText
- buttonEnabled
- variantExists
- hasVariants

### DIFF
--- a/test/unit/product/product-component.js
+++ b/test/unit/product/product-component.js
@@ -928,49 +928,56 @@ describe('Product Component class', () => {
 
     describe('getters', () => {
       describe('shouldUpdateImage', () => {
+        beforeEach(async () => {
+          await product.init(testProductCopy);
+        });
+
         it('returns true if there is no cached image', () => {
           product.cachedImage = null;
           assert.isTrue(product.shouldUpdateImage);
         });
 
-        it('returns true if image and cached image are different', async () => {
-          product.config.product.width = '100px';
-          await product.init(testProductCopy);
+        it('returns true if image and cached image are different', () => {
+          product = Object.defineProperty(product, 'image', {
+            value: {src: 'hat.jpg'},
+          });
           product.cachedImage = 'bar.jpg';
           assert.isTrue(product.shouldUpdateImage);
         });
 
-        it('returns false if image and cached image are the same', async () => {
-          product.config.product.width = '240px';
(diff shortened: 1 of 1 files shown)


test/unit/product/product-component.js
```diff
@@ -761,52 +761,50 @@ describe('Product Component class', () => {
...
         });
 
-        describe('if variant does not exist', () => {
-          it('returns cached image', async () => {
-            await product.init(testProductCopy);
-            product.selectedVariant = {};
-            assert.equal(product.currentImage.src, `${rootImageURI}image-one_280x420.jpg`);
+        it('does not set cached image to product.image if image should not update', async () => {
+          product = Object.defineProperty(product, 'shouldUpateImage', {
+            value: true,
```
melissaluu: Should this not be `false`? How is this test passing?

test/unit/product/product-component.js
```diff
@@ -761,52 +761,50 @@ describe('Product Component class', () => {
...
+          product.cachedImage = {
+            src: 'not-image-src',
+          };
+          await product.init(testProductCopy);
+          assert.notEqual(product.cachedImage, product.image);
+        });
+
+        it('returns cached image', async () => {
+          await product.init(testProductCopy);
+          assert.equal(product.currentImage, product.cachedImage);
```
melissaluu: Can you make this test more explicit on why the `product.currentImage` is equal to `product.cachedImage`?

test/unit/product/product-component.js
```diff
@@ -1043,113 +1048,230 @@ describe('Product Component class', () => {
...
         });
 
-        describe('if variant does not exist for selected options', () => {
-          it('returns false', async () => {
-            await product.init(testProductCopy);
-            product.selectedVariant = null;
-            assert.isNotOk(product.variantExists);
-          });
+        it('returns true if variant exists in model', async () => {
+          product.selectedVariant = {id: 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8xMjM0NQ=='};
```
melissaluu: Assuming this id string is coming from the fixture, are we able to reference the fixture value instead of this string? That will make it easier to understand where this string is coming from.

Issue #369, opened 2017-03-21
Expose image alt text


AmeliaBR:
The image templates should check for alt text & include it if available. 

If it is currently exposed in the data object, I couldn't figure out the variable names.

robskrob:
@AmeliaBR did you ever find a solution to this?

AmeliaBR:
@robskrob No, I didn't. Now that I check, the alt text is definitely there in the GraphQL responses that are coming through from the new API. But I'd have to do some digging to see if they ever get exposed in a way that the templates can access. The default templates are definitely not using it.

robskrob:
@AmeliaBR thank you for letting me know. I think you are right. 
> The default templates are definitely not using it.

I think [this is](https://github.com/Shopify/buy-button-js/blob/master/src/components/product.js#L218-L221 ) where the data for the carousel image is defined and no `alt text` is set.

AmeliaBR:
Searching this repository, I can't find any reference to the `altText` data field at all. The underlying SDK uses it in the request for image data, but I can't figure out all the layers of abstractions between that response and the data objects that are exposed to the templates.

I'm _pretty_ sure that the relevant connection is the `get image()` method in component/product.js, but there may be other places like the carousel image code that are currently destructuring data and only passing along select named properties.

robskrob:
@AmeliaBR thank you for sharing this [research](https://github.com/Shopify/buy-button-js/issues/369#issuecomment-447491225). I have created a [PR](https://github.com/Shopify/buy-button-js/pull/583) and I have tested my work locally -- I now see the alt text in the PDP images. Hopefully the PR could get reviewed quickly so I can respond to any feedback and so ultimately the PR can get merged. I'd really like to have alt text available on my images.

@AmeliaBR Let me know what you think with regards to the above PR I set up.

robskrob:
@spencercanner I've only interacted with you so far on this project. Would you be able to review [this PR](https://github.com/Shopify/buy-button-js/pull/583)? 

CC: @tessalt, @tanema, @harisaurus

AmeliaBR:
Thanks for doing this @robskrob !  I started to make the same changes myself, but realized it was too complex to do it without testing it out fully.

I left a comment with regards to the image carousel code. I've never used that particular template, so I'd never looked at it closely. The current code is really bad for accessibility. I gave suggestions for a bare-minimum fix and for more complicated options if you are willing to track down all the pieces.

melissaluu:
[...]

PR #583, opened 2018-12-14
add alt text to images

robskrob:
**Description**
As a developer
I want the alt text variable exposed on my Product Detail Page (PDP) images
So that I can properly set the alt text data, which exists in shopify, on the `alt` attribute of the main image and `aria-label` attribute on the carousel images.

**Acceptance Criteria**
* the `alt` and `aria-label` attribute data comes from the image alt text field in the shopify store
* the `aria-label` attribute is set the selected carousel image
* the `aria-label` attribute is set on the carousel image
* the `alt` attribute is set on the main image

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -144,7 +144,7 @@ export default class Product extends Component {
       src = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptionsLarge);
       srcOriginal = this.selectedImage.src;
-      altText = this.selectedImage.altText;
+      altText = (this.selectedImage.altText || this.model.title);
     } else if (this.selectedVariant.image == null && this.model.images[0] == null) {
       id = null;
       src = '';
@@ -156,14 +156,15 @@ export default class Product extends Component {
       src = this.model.images[0].src;
       srcLarge = this.props.client.image.helpers.imageForSize(this.model.images[0], imageOptionsLarge);
       srcOriginal = this.model.images[0].src;
-      altText = this.model.images[0].altText;
+      altText = (this.model.images[0].altText || this.model.title);
     } else {
       id = this.selectedVariant.image.id;
       src = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptionsLarge);
       srcOriginal = this.selectedVariant.image.src;
-      altText = this.selectedVariant.image.altText;
+      altText = (this.selectedVariant.image.altText || this.model.title);
     }
+
     return {id, src, srcLarge, srcOriginal, altText};
   }
 
@@ -218,13 +219,15 @@ export default class Product extends Component {
   }
 
   get carouselImages() {
+    let productTitle = this.model.title;
+
     return this.model.images.map((image) => {
       return {
         id: image.id,
         src: image.src,
         carouselSrc: this.props.client.image.helpers.imageForSize(image, {maxWidth: 100, maxHeight: 100}),
         isSelected: image.id === this.currentImage.id,
-        altText: image.altText,
+        altText: (image.altText || productTitle),
       };
     });
   }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -774,9 +774,8 @@ export default class Product extends Component {
     return model;
   }
 
-
-  imageAltText(string) {
-    return (string || this.model.title);
+  imageAltText(altText) {
+    return altText || this.model.title;
   }
 
 }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -144,7 +144,7 @@ export default class Product extends Component {
       src = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptionsLarge);
       srcOriginal = this.selectedImage.src;
-      altText = (this.selectedImage.altText || this.model.title);
+      altText = this.imageAltText(this.selectedImage.altText);
     } else if (this.selectedVariant.image == null && this.model.images[0] == null) {
       id = null;
       src = '';
@@ -156,13 +156,13 @@ export default class Product extends Component {
       src = this.model.images[0].src;
       srcLarge = this.props.client.image.helpers.imageForSize(this.model.images[0], imageOptionsLarge);
       srcOriginal = this.model.images[0].src;
-      altText = (this.model.images[0].altText || this.model.title);
+      altText = this.imageAltText(this.model.images[0].altText);
     } else {
       id = this.selectedVariant.image.id;
       src = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptionsLarge);
       srcOriginal = this.selectedVariant.image.src;
-      altText = (this.selectedVariant.image.altText || this.model.title);
+      altText = this.imageAltText(this.selectedVariant.image.altText);
     }
 
     return {id, src, srcLarge, srcOriginal, altText};
@@ -219,15 +219,13 @@ export default class Product extends Component {
   }
 
   get carouselImages() {
-    let productTitle = this.model.title;
-
     return this.model.images.map((image) => {
       return {
         id: image.id,
         src: image.src,
         carouselSrc: this.props.client.image.helpers.imageForSize(image, {maxWidth: 100, maxHeight: 100}),
         isSelected: image.id === this.currentImage.id,
-        altText: (image.altText || productTitle),
+        altText: this.imageAltText(image.altText),
       };
     });
   }
@@ -775,4 +773,10 @@ export default class Product extends Component {
     this.selectedVariant = selectedVariant;
     return model;
   }
+
+
+  imageAltText(string) {
+    return (string || this.model.title);
+  }
+
 }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -127,6 +127,7 @@ export default class Product extends Component {
     let src;
     let srcLarge;
     let srcOriginal;
+    let altText;
 
     const imageOptions = {
       maxWidth: imageSize,
@@ -143,23 +144,27 @@ export default class Product extends Component {
       src = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedImage, imageOptionsLarge);
       srcOriginal = this.selectedImage.src;
+      altText = this.selectedImage.altText;
     } else if (this.selectedVariant.image == null && this.model.images[0] == null) {
       id = null;
       src = '';
       srcLarge = '';
       srcOriginal = '';
+      altText = '';
     } else if (this.selectedVariant.image == null) {
       id = this.model.images[0].id;
       src = this.model.images[0].src;
       srcLarge = this.props.client.image.helpers.imageForSize(this.model.images[0], imageOptionsLarge);
       srcOriginal = this.model.images[0].src;
+      altText = this.model.images[0].altText;
     } else {
       id = this.selectedVariant.image.id;
       src = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptions);
       srcLarge = this.props.client.image.helpers.imageForSize(this.selectedVariant.image, imageOptionsLarge);
       srcOriginal = this.selectedVariant.image.src;
+      altText = this.selectedVariant.image.altText;
     }
-    return {id, src, srcLarge, srcOriginal};
+    return {id, src, srcLarge, srcOriginal, altText};
   }
 
   /**
@@ -219,6 +224,7 @@ export default class Product extends Component {
         src: image.src,
         carouselSrc: this.props.client.image.helpers.imageForSize(image, {maxWidth: 100, maxHeight: 100}),
         isSelected: image.id === this.currentImage.id,
+        altText: image.altText,
       };
     });
   }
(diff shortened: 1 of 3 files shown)


robskrob:
Resolves: https://github.com/Shopify/buy-button-js/issues/369

src/templates/product.js
```diff
@@ -19,15 +19,15 @@ const productTemplate = {
...
+                        <img class="{{data.classes.product.img}}" alt="{{data.currentImage.altText}}" src="{{data.currentImage.src}}" data-element="product.img" />
                         <button type="button" class="carousel-button carousel-button--next">
                           Right
                           <img class="carousel-button-arrow" src="//sdks.shopifycdn.com/buy-button/latest/arrow.svg" alt="Carousel Arrow"/>
                         </button>
                       </div>
                       <div class="{{data.classes.product.carousel}}">
                         {{#data.carouselImages}}
-                        <a data-element="product.carouselitem" href="{{src}}" class="{{data.classes.product.carouselItem}} {{#isSelected}} {{data.classes.product.carouselItemSelected}} {{/isSelected}}" data-image-id="{{id}}" style="background-image: url({{carouselSrc}})"></a>
+                        <a data-element="product.carouselitem" alt="{{altText}}" href="{{src}}" class="{{data.classes.product.carouselItem}} {{#isSelected}} {{data.classes.product.carouselItemSelected}} {{/isSelected}}" data-image-id="{{id}}" style="background-image: url({{carouselSrc}})"></a>
```
AmeliaBR: The `alt` attribute doesn't have an impact when used on an `<a>` element.

I'm not sure why the code is using `background-image` instead of a real `` element child. Maybe so the image won't load until that item is selected? But the original code is really, really bad accessibility — a link with no text content at all!

The simplest fix is to use an `aria-label` attribute to give the link an accessible name. Slightly better would be to use actual text content inside the `<a></a>`, and then hide it offscreen, but you'd have to dig into the SCSS to figure out what classes would do that.

But using the full image alt text as the link text is sub-optimal, since the purpose of clicking the link is to display the image (and therefore its full alt text). And equally important, it leaves the link label-less if there is no alt text in the database.

I'd recommend using something like "Image 1", "Image 2", "Image 3". Boring, but clear. Ideally, "Image" would be easily localizable using one of the text input configs.  (Same for "left" and "right" on the buttons, while we're at it.)

And since Mustache doesn't provide loop index variables, numbering the images would mean adding an `index` variable in the map created by the [`carouselImages` getter](https://github.com/Shopify/buy-button-js/blob/99d186ff641ea993401431c938b140126769ba5f/src/components/product.js#L215):

```js
  get carouselImages() {
    return this.model.images.map((image, index) => {
      return {
        index,
        id: image.id,
        src: image.src,
        carouselSrc: this.props.client.image.helpers.imageForSize(image, {maxWidth: 100, maxHeight: 100}),
        isSelected: image.id === this.currentImage.id,
      };
    });
  }
```

Anyway, that's a lot of different ideas. The bare minimum fix is to use `aria-label` instead of `alt`:

```suggestion
                        <a data-element="product.carouselitem" aria-label="{{altText}}" href="{{src}}" class="{{data.classes.product.carouselItem}} {{#isSelected}} {{data.classes.product.carouselItemSelected}} {{/isSelected}}" data-image-id="{{id}}" style="background-image: url({{carouselSrc}})"></a>
```

robskrob: @AmeliaBR Thank you for this review. I think I see what you mean with regards to your points on:
1. the decision to choose `<a>` over `` to render the image. Using an `` for the carousel images seems more intuitive.
2. the tradeoffs for using the alt text as the link text -- if alt text is available add it and hide it with SCSS but if alt text is missing then the link text would be blank, which is "sub-optimal".
3. "...using something like 'Image 1', 'Image 2', 'Image 3'" -- this is an interesting solution. I suppose if we were to insert `Image ${index}` as the link text then we would have to use SCSS to hide this text. However at the same time we would not need to worry about blank link text. 
4. "The bare minimum fix is to use aria-label instead of alt" -- Agreed 👍.

And in honor of 'keeping it simple' I'd like to go with option `4`. I'll make those changes now.

robskrob: @AmeliaBR Also what do you mean by:
> But the original code is really, really bad accessibility...

AmeliaBR: The really bad accessibility was having links with absolutely zero text content or aria-label. Using background images instead of `` is minor compared to that.

robskrob: Ok I think I understand. I am relatively new to the [Accessibility API](https://www.w3.org/WAI/PF/aria-implementation/#def_accessibility_api) so I think I got a little confused by the technical intent of the word, `accessibility`. An `aria-label` will make the `<a>` 'accessible' -- where as an `<a>` without text content or an aria-label makes the element unaccessible.

robskrob:
@tanema thank you for your feedback and for including another maintainer of this project on this PR!

src/templates/product.js
```diff
@@ -19,15 +19,15 @@ const productTemplate = {
                           Left
                           <img class="carousel-button-arrow" src="//sdks.shopifycdn.com/buy-button/latest/arrow.svg" alt="Carousel Arrow"/>
                         </button>
-                        <img class="{{data.classes.product.img}}" src="{{data.currentImage.src}}" data-element="product.img" />
+                        <img class="{{data.classes.product.img}}" alt="{{data.currentImage.altText}}" src="{{data.currentImage.src}}" data-element="product.img" />
```
spencercanner: We should probably be adding this alt text to the `img` product template as well

robskrob:
@spencercanner thank you for your feedback. This makes sense:
>  some images will end up with an empty alt tag
The way I would like to do this, personally, is I'd like to make an `image` model so that I can add methods on it like, `image#getAltTextMarkup`. Right now there appears to be just one large Product model `export default class Product extends Component {...` that encapsulates image logic, which should probably be defined, isolated, in its own domain or class. 

Instead of an isolated `image` model we have [this function](https://github.com/Shopify/buy-button-js/blob/a6392f26656039b7cce820bace5a90112de4c463/src/components/product.js#L220), which creates image objects. 

I like your suggestion here:
> In Shopify's online store themes, the alt text defaults to the product title if nothing has been set up in admin

Taking your suggestion I will update `get carouselImages() {` to set alt text with the product title if `altText` is falsey. While this does not create the modeling that I outline above, it is a smaller change which prevents the possibility of an empty `alt` and `aria-label`.

By this comment:
> We should probably be adding this alt text to the img product template as well

Do you mean to ask me to add `altText` to [this image element](https://github.com/Shopify/buy-button-js/blob/a6392f26656039b7cce820bace5a90112de4c463/src/templates/product.js#L15)?

robskrob:
@spencercanner I did find a bizarre use case however when testing my work. For some reason when I add `alt` text to the `img` element [here](https://github.com/Shopify/buy-button-js/blob/a6392f26656039b7cce820bace5a90112de4c463/src/templates/product.js#L15),  one of my images has a blank `alt` attribute `<img alt="{{data.currentImage.altText}}"...` -- which is the very thing we wanted to avoid. I added some defensive coding [here ](https://github.com/Shopify/buy-button-js/pull/583/files#diff-f9b3b5e78ec169f911886163beb184a7R147), [here](https://github.com/Shopify/buy-button-js/pull/583/files#diff-f9b3b5e78ec169f911886163beb184a7R159) and [here](https://github.com/Shopify/buy-button-js/pull/583/files#diff-f9b3b5e78ec169f911886163beb184a7R165) which for that image set the `alt` text attribute. However, I am not overjoyed with those defensive coding lines. I think this highlights the need for an `image` model so data validation can be secured. But as I was mentioning in [my earlier comment](https://github.com/Shopify/buy-button-js/pull/583#issuecomment-448276229), I think that would introduce a lot of change to the over all system, and I do not think at this current moment I am prepared to commit to that large refactor.

spencercanner:
@robskrob In buy-button-js, the models usually come from the js-buy-sdk directly so it might be better to create an image component, similar to the toggle. However, as you said, that refactor would require some significant thought and effort. In the meantime, we could abstract the `||` check into a helper function within the product class to avoid duplication. 

Alternatively, similar to `imageForSize`, this helper could live in the [js-buy-sdk](https://github.com/Shopify/js-buy-sdk/blob/master/src/image-helpers.js), since determining non-empty alt text seems like it could be a relatively common issue. It might be overkill for this simple check though.

robskrob:
@spencercanner thank you for your feedback. I should have time this weekend to work on this. Maybe I can get to it earlier than that.

robskrob:
@spencercanner 
>  In buy-button-js, the models usually come from the js-buy-sdk directly
I sorry for any confusion -- yes I think I understand what you mean by the models originating from `js-buy-sdk`. By model, I was just asking for an abstraction that could consolidate data into something more manageable and easier to reason about. It looks like we use components for this in `buy-button-js`, and I think, like you said, an image component would be appropriate to consolidate all the data and logic around images instead of having [product component](https://github.com/Shopify/buy-button-js/blob/master/src/components/product.js#L60) manage image data. And yes as you mentioned, 
> that refactor would require some significant thought and effort.

So I have [implemented](https://github.com/Shopify/buy-button-js/pull/583/files#diff-f9b3b5e78ec169f911886163beb184a7R778) what I believe reflects your suggestion:
>  we could abstract the || check into a helper function within the product class

robskrob:
@spencercanner just following up on my work for this PR. How does it look?

src/components/product.js
```diff
@@ -766,4 +773,10 @@ export default class Product extends Component {
     this.selectedVariant = selectedVariant;
     return model;
   }
+
+
+  imageAltText(string) {
```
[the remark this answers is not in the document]
robskrob: @spencercanner are the `get image()` tests [here](https://github.com/Shopify/buy-button-js/blob/master/test/unit/product/product-component.js#L1145)?

test/unit/product/product-component.js
```diff
@@ -1236,6 +1241,7 @@ describe('Product Component class', () => {
               src: firstImage.src,
               srcLarge: expectedSrcLarge,
               srcOriginal: firstImage.src,
+              altText: (firstImage.altText || 'test'),
```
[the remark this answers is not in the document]
robskrob: I removed the `||` and left, `'test'`. A bit literal but I always think thats good to do when testing. Moreover, I found `firstImage.altText` was null so the tests would fail if I chose `firstImage.altText`.

test/unit/product/product-component.js
```diff
@@ -1187,6 +1187,14 @@ describe('Product Component class', () => {
           assert.equal(image.srcLarge.maxHeight, image.srcLarge.maxWidth * 1.5);
         });
 
+        it('outputs the images alt text value when alt text is passed', () => {
```
spencercanner: Could we update this test name to `returns the passed in image alt text if it is valid` for consistency with other test names, and to reflect how the function is used?

src/components/product.js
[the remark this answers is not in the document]
spencercanner: We favour mocking out internal function calls where possible, as this allows us to test functions in isolation. Although `Product#imageAltText` is a simple function, we'd rather have it mocked out in `get image()` tests for consistency. Another benefit of doing so is that if the behaviour of `Product#imageAltText` were to change, we would only need to update its tests and not have to potentially update all the image related tests. Here's an [example](https://github.com/Shopify/buy-button-js/blob/master/test/unit/product/product-component.js#L611) of a test block that mocks out an internal function.

robskrob:
@spencercanner thank you for your feedback on the tests. I will do my best to get to this before the end of this Sunday (1/20/19). Just so I know, are the `get image()` tests [here](https://github.com/Shopify/buy-button-js/blob/master/test/unit/product/product-component.js#L1145)?

spencercanner:
@robskrob yeah, they're all located in that describe block.

src/components/product.js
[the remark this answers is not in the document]
robskrob: @spencercanner I'm really sorry but it is a bit unclear to me how I should proceed with regards to stubbing out this private, `imageForSize`, method. In my testing experience which is primarily with Ruby and Rspec, I would stub a `dependency` like an API call and typically let the private methods run on their own accord -- especially if they are just doing some string manipulation. If you want to step into this PR and enforce how you want the code to look like with regards to this stub you have in mind, please be my guest. 

I did, however, make an attempt to stub out `imageForSize` as I think you suggested, but I run into some errors:

Here's how I would stub:
```javascript
 describe('image', () => {
        let imageForSizeStub;
        let imageAltTextStub;

        beforeEach(async () => {
          await product.init(testProductCopy);
          imageForSizeStub = sinon.stub(product.props.client.image.helpers, 'imageForSize').callsFake((image, dimensions) => {
            return dimensions;
          });

          imageAltTextStub = sinon.stub(product, 'imageAltText');
        });

        afterEach(() => {
          imageForSizeStub.restore();
          imageAltTextStub.restore();
        });
```

and here are the errors:
```
1. returns object with id, src, srcLarge, and srcOriginal from selected image if selected image exists
2. returns object with id, src, srcLarge, and srcOriginal from first image in model if selected variant does not have an image
3. returns object with id, src, srcLarge, and srcOriginal from selected variant as default
```
The tests are complaining that the `expectedObject` should contain an `altText` value that matches the key value pair of `product.image`, which of course it does not because I stubbed out the private method `imageAltText`. I think in order to stub successfully I would need to stub out `imageAltText` to return whatever value the test expects for `altText` in each of the above tests.

Having said all that, I have added the tests in a `describe` block for `imageAltText()` so we do have some test coverage on that private method.

robskrob:
@spencercanner I made some updates but I struggled with the stub as reflected here in [this comment](https://github.com/Shopify/buy-button-js/pull/583#discussion_r249259693).

robskrob:
@spencercanner any time! I look forward to contributing more PRs to this project. It was a lot of fun working on it. When the follow up PR is ready with regards to stubbing feel free to ping me on github. I'd love to checkout what you make.

robskrob:
@spencercanner Thank you for merging this PR. I do not see the accessibility work of this PR in the latest version listed here: https://www.npmjs.com/package/@shopify/buy-button-js -- do you know when a version will be minted to contain this PRs work?

spencercanner:
@robskrob we're planning on doing a release early next week to propagate this work to all live buy buttons. I'll let you know once we've released it

spencercanner:
Hey @robskrob, just wanted to let you know that we've just released your work in version [0.10.0](https://www.npmjs.com/package/@shopify/buy-button-js/v/0.10.0) of buy-button-js. Thanks again for the contribution, we really appreciate it!

PR #585, opened 2018-12-17
Update product.js view tests

jameskim08:
Updated every getter and method in product.js within the view folder

### DIFF
--- a/test/unit/product/product-view.js
+++ b/test/unit/product/product-view.js
@@ -3,6 +3,7 @@ import Cart from '../../../src/components/cart';
 import ShopifyBuy from '../../../src/buybutton';
 import shopFixture from '../../fixtures/shop-info';
 import productFixture from '../../fixtures/product-fixture';
+import View from '../../../src/view';
 
 const config = {
   id: 123,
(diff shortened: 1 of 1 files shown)


test/unit/product/product-view.js
```diff
@@ -54,40 +55,210 @@ describe('Product View class', () => {
...
+      product = Object.defineProperty(product, 'isButton', {
+        writable: true,
+      });
     });
 
-    describe('when isButton() is false', () => {
-      it('calls super', () => {
+    it('wraps html in a div with wrapper class and product class if the component is not a button', () => {
+      product.isButton = false;
+      const htmlString = product.view.wrapTemplate('test');
```
spencercanner: For consistency with other tests, can we update this variable name to `string`, or update everything to `htmlString`?

test/unit/product/product-view.js
```diff
@@ -54,40 +55,210 @@ describe('Product View class', () => {
...
+      product.config.product.contents = {
+        img: null,
+        imgWithCarousel: null,
+      };
+      product.view.resizeOnLoad();
+      assert.notCalled(addEventListenerStub);
+    });
+
+    it('does not add event listener if there is no image in the wrapper', () => {
+      product.config.product.contents = {
```
spencercanner: For clarity, could we move the instantiation of `node` and `wrapper.appendChild` out of the `beforeEach` and into the tests below? We could explicitly add classes to `node` which either match or don't match `img-class`.

test/unit/product/product-view.js
```diff
@@ -54,40 +55,221 @@ describe('Product View class', () => {
...
+    });
+
+    afterEach(() => {
+      resizeStub.restore();
+      addEventListenerStub.restore();
+    });
+
+    it('does not add event listener if product contents do not have an image or image with carousel', () => {
+      const wrapper = document.createElement('div');
+      product.view.wrapper = wrapper;
```
spencercanner: Could we leave 
```
const wrapper = document.createElement('div');
product.view.wrapper = wrapper;
```
in the beforeEach?

PR #586, opened 2018-12-17
Updated test in product.js in updater

jameskim08:
Refactored and created tests in updateConfig for product.

### DIFF
--- a/test/unit/product/product-updater.js
+++ b/test/unit/product/product-updater.js
@@ -62,7 +62,7 @@ describe('Product Updater class', () => {
   });
 
   describe('updateConfig()', () => {
-    describe('if storefrontId or storefrontVariantId is supplied from config', () => {
+    describe('if id or variant id is supplied in config', () => {
       it('calls init and returns if id is updated', () => {
         product.updateConfig({id: 123});
         assert.calledOnce(initStub);
--- a/test/unit/product/product-updater.js
+++ b/test/unit/product/product-updater.js
@@ -1,6 +1,6 @@
 import Product from '../../../src/components/product';
+import Updater from '../../../src/updater';
 import Cart from '../../../src/components/cart';
-import Modal from '../../../src/components/modal';
 import ShopifyBuy from '../../../src/buybutton';
 import shopFixture from '../../fixtures/shop-info';
 import productFixture from '../../fixtures/product-fixture';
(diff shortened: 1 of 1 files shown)


test/unit/product/product-updater.js
```diff
@@ -34,108 +38,216 @@ describe('Product Updater class', () => {
...
-
-    describe('when updating ID, storefront ID, variant ID, or storefront variant ID', () => {
-      let initSpy;
-
-      beforeEach(() => {
-        initSpy = sinon.spy(product, 'init');
-      });
-
-      it('calls init if ID updated', () => {
+    describe('when updating id, storefront id, variant id, or storefront variant id', () => {
```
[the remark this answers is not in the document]
spencercanner: It seems like there's a lot of overlap between this describe title, and the individual test names. Maybe we could remove this describe, or rename it to reflect what happens in the `if` statement?

test/unit/product/product-updater.js
```diff
@@ -34,108 +38,216 @@ describe('Product Updater class', () => {
...
+        updateConfig: updateConfigSpy,
+      };
+      const expectedObject = {
+        options: JSON.parse(JSON.stringify(product.config)),
+      };
+      expectedObject.options.product = product.modalProductConfig;
+
+      product.updateConfig(newConfig);
+      assert.calledOnce(updateConfigSpy);
+      assert.calledWith(updateConfigSpy, expectedObject);
```
spencercanner: Is there a way to check that it's also using the config passed into `Product.updateConfig`?

test/unit/product/product-updater.js
```diff
@@ -34,108 +39,218 @@ describe('Product Updater class', () => {
...
-
-    describe('when updating ID, storefront ID, variant ID, or storefront variant ID', () => {
-      let initSpy;
-
-      beforeEach(() => {
-        initSpy = sinon.spy(product, 'init');
-      });
-
-      it('calls init if ID updated', () => {
+    describe('if storefrontId or storefrontVariantId is supplied from config', () => {
```
spencercanner: I think technically normal ids can be specified in the config, the `normalizeConfig` just transforms them into their respective storefront version. Maybe we could say `if id or variant id is supplied in config`, to be generic enough to encompass normal and storefront ids

PR #588, opened 2018-12-17
Updated tests for checkout component tests

jameskim08:
This was not put into a folder like the other children classes because checkout only has component and does not have a view or updater class.

Also, open does not check the location of window because it is currently not possible

### DIFF
--- a/test/unit/checkout.js
+++ b/test/unit/checkout.js
@@ -1,11 +1,55 @@
 import Checkout from '../../src/components/checkout';
 
 describe('Checkout class', () => {
-  describe('get params', () => {
-    it('puts together a big param string', () => {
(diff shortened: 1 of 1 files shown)


test/unit/checkout.js
```diff
@@ -1,11 +1,55 @@
...
+        };
+        checkout.open('test.com');
+        assert.calledOnce(openStub);
+        assert.calledWith(openStub, 'test.com', 'checkout', checkout.params);
+      });
+    });
+
+    describe('getters', () => {
+      describe('params', () => {
+        it('creates a string including window config params', () => {
```
spencercanner: Maybe update getter test names to `it('returns...`, for consistency with other test files

PR #589, opened 2018-12-18
Created Tests for Toggle Component

jameskim08:
Deleted toggle.js test file
Created folder for toggle, similarly to product
Created toggle component tests in toggle-component.js


test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,151 @@
+import Toggle from '../../../src/components/toggle';
+import ToggleView from '../../../src/views/toggle';
+
+const fakeClient = {
```
spencercanner: Do we need the fakeClient? The toggle component doesn't seem to be accessing the client from the props.

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,151 @@
...
+      document.body.removeChild(cart.node);
+    });
+
+    it('sets typeKey to toggle', () => {
+      toggle = new Toggle({}, {client: fakeClient, cart});
+      assert.equal(toggle.typeKey, 'toggle');
+    });
+
+    it('sets node to node in config if provided', () => {
+      const node = {};
```
spencercanner: Could we set this to a `document.createElement` to more closely mimic a real node?

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,151 @@
...
+      assert.calledOnce(insertBeforeStub);
+      assert.calledWith(insertBeforeStub, div, cart.node);
+      assert.equal(toggle.node, node);
+
+      insertBeforeStub.restore();
+      createElementStub.restore();
+    });
+
+    it('creates a new toggle view', () => {
+      toggle = new Toggle({}, {client: fakeClient, cart});
```
spencercanner: Similar to above, this test might be leaving a residual `div` in the DOM.

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,151 @@
...
+      };
+      document.body.appendChild(cart.node);
+    });
+
+    afterEach(() => {
+      document.body.removeChild(cart.node);
+    });
+
+    it('sets typeKey to toggle', () => {
+      toggle = new Toggle({}, {client: fakeClient, cart});
```
spencercanner: Since we're not passing in a node or stubbing out the document methods, I think this test might be leaving a residual div in the DOM

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,151 @@
...
+            },
+            {
+              id: 2,
+              quantity: 2,
+            },
+          ],
+        },
+      };
+      document.body.appendChild(cart.node);
+      toggle = new Toggle({}, {client: fakeClient, cart});
```
spencercanner: Similar to above, since not stubbing out the document methods, the constructor's `insertBefore` is going to be adding divs to the actual DOM which aren't cleaned up.

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,147 @@
...
+  let node;
+
+  beforeEach(() => {
+    node = document.createElement('div');
+  });
+
+  describe('constructor', () => {
+    beforeEach(() => {
+      cart = {
+        node: document.createElement('div'),
```
spencercanner: Since we're passing in `node` for all tests but one, we can probably move the contents of the beforeEach and afterEach into that test

test/unit/toggle/toggle-component.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+  });
+
+  describe('prototype methods', () => {
+    let toggleVisibilitySpy;
+
+    beforeEach(() => {
+      toggleVisibilitySpy = sinon.spy();
+      cart = {
+        node: document.createElement('div'),
```
spencercanner: Similar to above, I think we can remove this as well as the `appendChild` and `removeChild`, since we're passing in the node.

spencercanner: I think we could also remove the node from the cart, since it won't be used

PR #592, opened 2018-12-19
Created tests for modal component

jameskim08:
Similarly to toggle and product, modal was moved into its own folder which will separate view, component, and updater.


test/unit/modal/modal-component.js
```diff
@@ -0,0 +1,252 @@
...
+        });
+      });
+
+      describe('get productConfig', () => {
+        it('returns object that includes global config', () => {
+          assert.deepInclude(modal.productConfig, modal.globalConfig);
+        });
+
+        it('returns object with node that holds product wrapper', () => {
+          assert.deepInclude(modal.productConfig, {node: modal.productWrapper});
```
spencercanner: Can we set `productWrapper` to a `div` in this test, so it's not null?

PR #601, opened 2019-01-23
Add notes field to cart

spencercanner:
**What are you trying to accomplish with this PR?**
Expose a note field at the cart level, which will populate the checkout object's [note field](https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/object/checkout#note). 

I've added a new contents option called `note`, a new text field called `noteDescription`, as well as several classes which can be used to override the default styling. 
The styling of the cart note follows the existing convention of using padding around the scrollable area where the line items appear. However, this method does not allow for a long cart notice or note description text without cutting off some of the scrollable area.  This was the case before as well, for long cart notice strings. 

To 🎩 : 
* Check that the new config options work and allow you to customize the cart note display
* Check that the cart note persists if you enter some text, close your browser, and come back to the cart
* Proceed with a checkout (easiest way is to have $0 products), and check that the cart note appears on the order page in the admin

### DIFF
--- a/test/unit/cart.js
+++ b/test/unit/cart.js
@@ -11,8 +11,19 @@ import ShopifyBuy from '../../src/buybutton';
 let cart;
 
 describe('Cart class', () => {
+  let closeCartSpy;
+
   beforeEach(() => {
-    cart = new Cart({}, {
+    closeCartSpy = sinon.spy();
+    cart = new Cart({
+      options: {
+        cart: {
+          viewData: {
+            test: 'test string',
+          },
+        },
+      },
+    }, {
       client: ShopifyBuy.buildClient({
         domain: 'test.myshopify.com',
         storefrontAccessToken: 123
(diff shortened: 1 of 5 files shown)


test/unit/cart.js
```diff
@@ -11,8 +11,19 @@ import ShopifyBuy from '../../src/buybutton';
...
+
   beforeEach(() => {
-    cart = new Cart({}, {
+    closeCartSpy = sinon.spy();
+    cart = new Cart({
+      options: {
+        cart: {
+          viewData: {
+            test: 'test string',
+          },
```
melissaluu: For completeness, I think adding populated `contents` and `text` objects will better validate the `viewData` unit tests.

spencercanner:
Tested on the following browsers: 
- [ ] Chrome
- [ ] Safari 
- [ ] Firefox 
- [ ] Safari iOS
- [ ] Edge (Windows VM)
- [ ] IE11 (Windows VM)
- [ ] IE10 (Windows VM)

PR #640, opened 2019-08-19
Display automatic discounts in line item and cart subtotal

spencercanner:
**Line item**
* Add in a new line-item component template for `priceWithDiscount`, and set its contents flag to true in defaults
* Add new classes for `priceWithDiscount` template
* Set the line-item component's `price` contents flag to false by default (breaking change)
* Add logic to populate the new data fields for the template, pulling from the line item's `discountAllocations` field 
* Remove optimistic updating of quantity and price when updating quantity, and only display updated once they're returned from SF API
* Disable pointer events and reduce opacity on the quantity fields when updating the line item's quantity 
* Update the order of `title` and `variantTitle`, and update styling so the variant title appears below the product title

**Cart**
* Add `discounts` contents flag to cart component 
* Add discounts markup to the cart footer template
* Add new classes for `discounts` markup
  * For backwards compatibility, the `subtotal` related styles are being applied to the `discounts`  
* Add logic to populate the new data fields for the template pulling from the checkout's `discountApplications` field
* Fetch the cart subtotal based on the value of the `discounts` contents flag
* Add a new class to the cart's scrollable area, to better handle the additional footer height created by discounts


src/components/cart.js
```diff
@@ -78,10 +79,28 @@ export default class Cart extends Component {
   get lineItemsHtml() {
     return this.lineItemCache.reduce((acc, lineItem) => {
       const data = Object.assign({}, lineItem, this.options.viewData);
+      const fullPrice = data.variant.priceV2.amount * data.quantity;
+      const formattedPrice = formatMoney(fullPrice, this.moneyFormat);
+      const discountAllocations = data.discountAllocations;
+
+      let totalDiscount = 0;
```
melissaluu: Since you're using a reduce below - you could use the resulting "object" to store your total. This way the external `totalDiscount` won't update via a reduce function.

src/components/cart.js
```diff
@@ -78,10 +79,28 @@ export default class Cart extends Component {
...
+      const discounts = discountAllocations.reduce((discountArr, discount) => {
+        const targetSelection = discount.discountApplication.targetSelection;
+        if (targetSelection === 'ENTITLED' || targetSelection === 'EXPLICIT') {
```
melissaluu: Are these key words to indicate discounts for line items? If so, maybe a helper function or an array to store these keys would be helpful and more explicit.

src/components/cart.js
```diff
@@ -96,22 +115,56 @@ export default class Cart extends Component {
...
    */
   get formattedTotal() {
-    return formatMoney(this.model.subtotalPrice, this.moneyFormat);
+    const total = this.options.contents.discounts ? this.model.subtotalPriceV2.amount : this.model.lineItemsSubtotalPrice.amount;
+    return formatMoney(total, this.moneyFormat);
   }
 
-  get formattedLineItemsSubtotal() {
-    return formatMoney(this.model.lineItemsSubtotalPrice.amount, this.moneyFormat);
+  get discounts() {
```
melissaluu: Is this the cart discount? If so, maybe the discount variable names could be more explicit since it can get confusing between the different sets of discounts.

src/components/cart.js
```diff
@@ -96,22 +115,56 @@ export default class Cart extends Component {
...
+
+    return this.model.discountApplications.reduce((discountArr, discount) => {
+      if (discount.targetSelection === 'ALL') {
+        let discountValue = 0;
+        if (discount.value.amount) {
+          discountValue = discount.value.amount;
+        } else if (discount.value.percentage) {
+          discountValue = (discount.value.percentage / 100) * this.model.lineItemsSubtotalPrice.amount;
+        }
+        discountArr.push({text: discount.title, amount: `- ${formatMoney(discountValue, this.moneyFormat)}`});
```
melissaluu: Is there a discount that isn't amount or percentage? If so, should the `discountArr` only update if discount value is greater than zero?

src/components/cart.js
```diff
@@ -324,7 +377,13 @@ export default class Cart extends Component {
   updateItem(id, quantity) {
     this._userEvent('updateItemQuantity');
     const lineItem = {id, quantity};
-    this.updateCacheItem(id, quantity);
+    const lineItemEl = this.view.document.getElementById(id);
+    if (lineItemEl) {
+      const quantityEl = lineItemEl.getElementsByClassName(this.classes.lineItem.quantity)[0];
+      if (quantityEl) {
+        addClassToElement('is-loading', quantityEl);
```
melissaluu: Just a sanity check but after we get the new checkout model back, it'll trigger re-render and that would remove the loading state right?

src/components/cart.js
```diff
@@ -78,10 +82,30 @@ export default class Cart extends Component {
   get lineItemsHtml() {
     return this.lineItemCache.reduce((acc, lineItem) => {
       const data = Object.assign({}, lineItem, this.options.viewData);
+      const fullPrice = data.variant.priceV2.amount * data.quantity;
+      const formattedPrice = formatMoney(fullPrice, this.moneyFormat);
+      const discountAllocations = data.discountAllocations;
+
+      const {discounts, totalDiscount} = discountAllocations.reduce((discountAcc, discount) => {
+        const targetSelection = discount.discountApplication.targetSelection;
+        if (LINE_ITEM_TARGET_SELECTIONS.indexOf(targetSelection) > -1) {
```
melissaluu: Small nitpick - we could use `array.includes` here 😄

spencercanner: It's not supported natively in IE11 and Safari 8, and we weren't bringing in the core-js polyfill so I opted to used this instead. I don't really have a preference though

melissaluu: Ahh boo - I thought we were bringing in the array polyfills 🤔 
I'm not set on using `includes` (I just find it easier to read) but it might be worthwhile bringing in the array polyfills if we currently not (we can do that separately too I guess)

PR #641, opened 2019-08-20
Replace postcss-cssnext with postcss-preset-env

spencercanner:
* remove `postcss-cssnext` package, since it's deprecated
* add the suggested replacement package: `postcss-preset-env`
* update styles scripts to use the new package
  * using `stage: 1` which should provide coverage over all css features we're using - most notably nesting which was moved from stage 0 -> 1 (https://cssdb.org/#nesting-rules)
* add `browserslist` to package.json, which is what autoprefixer uses to determine which css properties require what prefixes


src/styles/embeds/sass/components/buttons.css
```diff
@@ -86,6 +86,6 @@
 
   &:hover {
     transform: scale(1.2);
-    color: color(var(--color-cart-title) lightness(- 5%));
+    color: color(var(--color-cart-title) lightness(-5%));
```
spencercanner: These values were causing a `ParserError` when running the style scripts

PR #680, opened 2020-04-16
Fixed issue #672 - multiple collections rendered in random DOMs regardless of client config when iframe for productSet is FALSE

### LINKED ISSUE #672: Creating two collection components with separate nodes combines the collections into the first node when productSet iframe=false.
I created two different divs, #collection-1 and #collection-2. I created two collection components, the first one specifying #collection-1 as it's node and the 2nd specifying #collection-2.

When I load the page I am seeing the products from component collection-2 getting added to the products from #collection-1. #collection-2 has the built out html for the collection, including a pagination button, but contains no products.

I expect the second collection component, where node #collection-2 is specified, to contain the products from the other collection.

The goal is to have tabs containing separate collections that have a shared cart.

SpareShade:
This should fix #672 - conditional check if productSet iframe is false

Added a conditional to ```productConfig``` ```node``` setting in   ```ProductSet``` ```renderProducts``` method 

```javascript
    const productConfig = Object.assign({}, this.globalConfig, {
      node: this.view.document.querySelector(`.${this.classes.productSet.products}`),
      options: merge({}, this.config, {
        product: {
          iframe: false,
          classes: {
            wrapper: this.classes.productSet.product,
          },
        },
      }),
    });

    // fix for #672 - code updated after the 'build fail'
    if (this.config.productSet.iframe === false) {
      productConfig.node = this.node.querySelector(`.${this.classes.productSet.products}`);
    }

```

### DIFF
--- a/src/components/product-set.js
+++ b/src/components/product-set.js
@@ -225,7 +225,6 @@ export default class ProductSet extends Component {
       }),
     });
 
-    // fix for #672
     if (this.config.productSet.iframe === false) {
       productConfig.node = this.node.querySelector(`.${this.classes.productSet.products}`);
     }
--- a/src/components/product-set.js
+++ b/src/components/product-set.js
@@ -227,8 +227,7 @@ export default class ProductSet extends Component {
 
     // fix for #672
     if (this.config.productSet.iframe === false) {
-        productConfig.node = this.node.querySelector(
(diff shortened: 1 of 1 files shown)


SpareShade:
Build failed, will update the code

PR #681, opened 2020-04-18
Fix CSS class for a product's disabled button

gabrielmaldi:
Currently, when a product's button is disabled (e.g. no stock), it's rendered with `undefined` as class:

```html
<button disabled="" class="shopify-buy__btn undefined " data-element="product.button"></button>
```

This patch fixes this by using the intended default class `shopify-buy__btn-disabled`:

```html
<button disabled="" class="shopify-buy__btn shopify-buy__btn-disabled " data-element="product.button"></button>
```

### DIFF
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -274,7 +274,7 @@ export default class Product extends Component {
   }
 
   get buttonClass() {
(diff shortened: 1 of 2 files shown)


gabrielmaldi:
My pleasure, @melissaluu 😊

PR #707, opened 2020-08-25
Add focus trap when cart or modal are open

spencercanner:
* Add a common utility to trap focus within a given container
  * This function also focuses the first element in the container by default
* Call focus trap function from existing `setFocus` function, which is used by both the cart and modal
  * The first element focusing was removed from `setFocus` since that is handled in the focus trap function
* Update modal and cart to remove the focus trap event listeners on close

To 🎩 : 
Cart:
* Open the cart
* Continuously tab forwards through the cart and verify that focus stays in the cart
* Continuously tab backwards through the cart and verify that focus stays in the cart
* Close the cart by hitting enter when focused on the close button
  * Focus will stay on the close button (will be fixed in a followup PR)
* Tab backwards and verify that other elements on the page can be focused
  * This is to test that the focus trap is removed when the cart is closed

Modal: 
* Open a product details modal
* Continuously tab forwards through the modal and verify that focus stays in the modal
* Continuously tab backwards through the modal and verify that focus stays in the modal
* Close the modal by hitting enter when focused on the close button
* Verify that focus is returned back to the `view product` button
* Tabbing forwards or backwards should focus on other page elements as expected

### DIFF
--- a/src/utils/focus.js
+++ b/src/utils/focus.js
@@ -1,0 +1,55 @@
+const TAB_KEY = 9;
+const trapFocusHandlers = {};
+
+export function removeTrapFocus(container) {
+  if (trapFocusHandlers.focusin) {
+    container.removeEventListener("focusin", trapFocusHandlers.focusin);
+  } 
+  if (trapFocusHandlers.focusout) {
+    container.removeEventListener("focusout", trapFocusHandlers.focusout);
+  }
+  if (trapFocusHandlers.keydown) {
+    container.removeEventListener("keydown", trapFocusHandlers.keydown);
+  }
+}
+
+export function trapFocus(container) {
+  const focusableElements = container.querySelectorAll("a, button:enabled, input:enabled, select:enabled");
+  const firstElement = focusableElements[0];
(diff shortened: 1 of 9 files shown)


src/utils/focus.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  };
+
+  container.addEventListener("focusout", trapFocusHandlers.focusout);
+  container.addEventListener("focusin", trapFocusHandlers.focusin);
+
+  firstElement.focus();
+}
+
+export function removeTrapFocus(container) {
```
melissaluu, line 45: Just a nit but since this function is called in `trapFocus`, does it make sense to move this up top?

Also, this is probably more defensive than it needs to be but if `trapFocusHandlers` is empty, we technically don't need to run the following code

spencercanner: I can add a check before each `removeEventListener` to make sure that `trapFocusHandlers` contains the listener function that is being removed
```
if (trapFocusHandlers.focusin) {
  container.removeEventListener("focusin", trapFocusHandlers.focusin);
} 
```

PR #709, opened 2020-08-27
Return focus to the right element when the cart is closed

spencercanner:
* Add a new property to the product component to track whether the product is rendered in the modal or not
* Use the new property to determine whether or not the last active element should be set to the button when the add to cart button is clicked
  * If the product is not a modal product, the active element should be set so focus is returned to the button
  * If the product is a modal product, the active element should not be set because the modal is closed when the button is clicked, so focus should not be returned to that button in the modal
* Set the active element to the toggle when the cart is opened via the toggle 
  * This had to be set in two places since the click and keydown events are handled separately
* Fix a double call to `setFocus` when the cart is opened
  * Previously there was a race condition where the focus would be returned to the cart by the second `setFocus` if the cart was closed too quickly after being opened

**To 🎩 :**
Product: 
* Create a buy button for a single product where the button adds to cart
* Tab to the button, and add the item to cart by pressing enter 
* Once the cart is open, press enter when focused on the close button
* Verify that the focus is returned to the product's `add to cart` button
* Click the add to cart button using the mouse, and close the cart using the mouse
* Verify that the focus is returned to the product's `add to cart` button

Collection/Product Set:
* Create a buy button for a collection/product set where the button adds to cart
* Tab to one of the product buttons, and add the item to the cart by pressing enter
* Once the cart is open, press enter when focused on the close button
* Verify that focus is returned to the `add to cart` button for the product that was added
* Click the add to cart button for one of the products using the mouse, and close the cart using the mouse
* Verify that the focus is returned to the `add to cart` button for the product that was added

Modal: 
* Create a buy button for a product where the button opens the product details modal
* Tab to the button, and open the modal by pressing enter
* Once the modal is open, tab through the modal to the add to cart button
* Add the item by pressing enter (the modal should close then the cart opens)
* Once the cart is open, press enter when focused on the close button
* Verify that the focus is returned to the product's `view details` button
* Open the product details modal using the mouse, and add the product to cart using the mouse
* Close the cart using the mouse 
* Verify that the focus is returned to the product's `view details` button

Toggle: 
* Tab to the cart toggle, and open the cart by pressing enter
* Once the cart is open, press enter when focused on the close button
* Verify that focus is returned to the cart toggle
* Click the toggle using the mouse, and close the cart using the mouse
* Verify that focus is returned to the cart toggle

### DIFF
--- a/src/components/cart.js
+++ b/src/components/cart.js
@@ -428,7 +428,9 @@ export default class Cart extends Component {
         this.updateCache(this.model.lineItems);
         this.view.render();
         this.toggles.forEach((toggle) => toggle.view.render());
-        this.view.setFocus();
+        if (!openCart) {
+          this.view.setFocus();
+        }
         return checkout;
       });
     } else {
@@ -443,7 +445,9 @@ export default class Cart extends Component {
         this.updateCache(this.model.lineItems);
         this.view.render();
         this.toggles.forEach((toggle) => toggle.view.render());
-        this.view.setFocus();
+        if (!openCart) {
+          this.view.setFocus();
+        }
         return checkout;
       });
     }
--- a/src/components/product.js
+++ b/src/components/product.js
@@ -82,6 +82,7 @@ export default class Product extends Component {
     this.selectedVariant = {};
     this.selectedOptions = {};
     this.selectedImage = null;
+    this.modalProduct = Boolean(config.modalProduct);
     this.updater = new ProductUpdater(this);
     this.view = new ProductView(this);
   }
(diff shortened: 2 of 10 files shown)


src/components/cart.js
```diff
@@ -428,7 +428,9 @@ export default class Cart extends Component {
         this.updateCache(this.model.lineItems);
         this.view.render();
         this.toggles.forEach((toggle) => toggle.view.render());
-        this.view.setFocus();
+        if (!openCart) {
```
spencercanner, line 431: If `openCart` is true, it triggers a call to `cart.open()` above, which calls `setFocus` as well

src/components/product.js
```diff
@@ -649,7 +650,7 @@ export default class Product extends Component {
       this.props.closeModal();
       this._userEvent('addVariantToCart');
       this.props.tracker.trackMethod(this.cart.addVariantToCart.bind(this), 'Update Cart', this.selectedVariantTrackingInfo)(this.selectedVariant, this.selectedQuantity);
-      if (this.iframe) {
+      if (!this.modalProduct) {
```
spencercanner, line 653: `this.iframe` is always undefined since there is no iframe property on the product component, so `setActiveEl` was never called. I also don't see a reason to skip focus tracking in non-iframe mode, so I think it's safe to remove the check

melissaluu (approved):

PR #710, opened 2020-08-31
Prevent focus on cart elements when closed

spencercanner:
* Add `visibility: hidden` to the cart wrapper by default, and set it to `visibility: visible` when the cart is open
* Add `is-visible` class to change the wrapper to `visibility: visible` when open
* Update the transition to only happen for the transform 
  * This ensures that in non-iframe mode, the cart is visible immediately so the close button can be focused
* Add `display: none` to the cart's iframe by default 
  * This is required on top of visibility hidden because some screen readers still pick up the iframe within a hidden container
* Add `is-block` class to change the iframe to `display: block` when the cart is opened
* Add a listener to the cart container's transition event so the display can be changed back to none and visibility back to hidden, once the cart has completed the transition to the closed state
* Wrap the call to `setFocus` in a timeout so the DOM class attribute updates finish before. This prevents focusing an element in a `display: none` or `visibility: hidden` container

* Move the cart over to the new folder test structure, and create a new `cart-view.js` test file

* Note: depending on the browser the cart toggle iframe may receive focus after the toggle, so an extra tab may be required after being focused on the toggle

--- 
**To 🎩 :** 

Toggle mouse click
* Click the cart toggle and verify that the cart still slides open
* Verify that the cart elements can still be interacted with
* Verify that the cart transitions out and closes when the `X` button is clicked
* Verify that the cart transitions out and closes when the area outside of the cart is clicked

Add to cart mouse click
* Click add to cart and verify that the cart still slides open
* Verify that the cart elements can still be interacted with
* Verify that the cart transitions out and closes when the `X` button is clicked
* Verify that the cart transitions out and closes when the area outside of the cart is clicked

Cart missing from tab order when closed
* Tab through the page
* Verify that the focus returns to the top of the browser after focusing out of the cart toggle / last element on the page (depending on the browser, you might need an extra tab to get to the top of the browser)

Toggle keyboard
* Tab through the page to the cart toggle
* Click the cart toggle by pressing the enter key 
* Verify that the cart slides out and focus is sent to the `X` button
* Verify that the cart elements can still be tabbed through 
* Verify that the cart transitions out and closes when pressing enter on the `X` button
* Reverse tab until you get to the cart toggle (this is necessary until #709 is shipped since the focus will remain on the cart close button)
* Verify that tabbing forward will send focus to the top of the browser, and not into the cart

Add to cart keyboard
* Tab through the page to an add to cart button
* Click the add to cart button by pressing the enter key 
[...]

PR #711, opened 2020-09-01
Prevent default event when triggering the cart toggle using the keydown event

spencercanner:
* Fix a bug in firefox and legacy edge which prevented the cart from being opened by pressing the enter key on the keyboard
* When pressing the enter key, a custom event handler would trigger the cart to open, but it would subsequently also trigger a click on the cart's close button, which became focused during the custom handler

To 🎩 : 
* Tab to the cart toggle 
* Verify that cart toggle can be clicked by pressing the enter key 
* Verify that the cart can be tabbed through
* Verify that clicking `X` button in the cart closes the cart

Browsers to 🎩 : 
- [x] Safari - Mac
- [x] Chrome - Mac 
- [x] Edge - Mac 
- [x] Firefox - Mac
- [x] Legacy Edge - Windows
- [x] Edge - Windows
- [x] Chrome - Windows 
- [x] Firefox - Windows

### DIFF
--- a/src/views/toggle.js
+++ b/src/views/toggle.js
@@ -57,6 +57,7 @@ export default class ToggleView extends View {
       if (evt.keyCode !== ENTER_KEY) {
         return;
       }
+      evt.preventDefault();
       this.component.props.cart.toggleVisibility(this.component.props.cart);
     });
   }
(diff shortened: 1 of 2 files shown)


spencercanner:
Change made in #710 to assign focus to the close button after all DOM class attribute mutations are completed, so this is no longer necessary

PR #792, opened 2021-12-08
[m1 compatibility] replace node-sass with sass 

kevcube:

PR #816, opened 2022-08-09
Update repo to use sass and latests JS Buy SDK version

melissaluu:
Update repo to use the latest `sass` instead of `node-sass`. Also update JS Buy SDK to the latest version.

Rel: #792


fightingtheboss (approved):
Looks simple enough and 🎩  was smooth!

PR #821, opened 2022-08-12
Remove encoding during normalizing the ids

melissaluu:
Remove the `btoa` encoding of `ids` since Storefront API does not return `base64` ids anymore. 

To 🎩 : 
- add a `variantId` value to the Buy Button config where the variant id is not the first/default variant

### DIFF
--- a/src/utils/normalize-config.js
+++ b/src/utils/normalize-config.js
@@ -1,5 +1,5 @@
 function normalizeId(type, databaseKey) {
-  return btoa(`gid://shopify/${type}/${databaseKey}`);
+  return `gid://shopify/${type}/${databaseKey}`;
 }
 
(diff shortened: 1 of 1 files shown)


fightingtheboss (approved):
Changes look good, tests passed for me!

I was curious why you changed some of the IDs in the tests, but my assumption was to make them a little more real than 1234.

melissaluu:
> I was curious why you changed some of the IDs in the tests, but my assumption was to make them a little more real than 1234.

The test fixtures uses simplified ids for values that's provided by the user via the config. For this PR, I only changed the ids that are provided by SFAPI - can definitely look into updating those ids to full values in the future.

frenbase:
Hi @melissaluu good day. 
Can I know what are the fast way to fix the latest buy-button-storefront.js on btoa() removal, as now my production site are broken, not able to accept checkout id and product id?

PR #894, opened 2025-04-10
3.0.1

juanpprieto:


wohali:
Hi @juanpprieto - this seems to be missing the `package.json` and `yarn.lock` changes from #888. It is only updating the README. Is that intentional?

PR #947, opened 2026-03-10
chore: add CODEOWNERS for developer-platforms team

kdaviduik:
Auto-request reviews from @Shopify/developer-platforms on all PRs,
matching the pattern used by Hydrogen and other team repos. Example:
https://github.com/Shopify/hydrogen/blob/main/.github/CODEOWNERS

PR #946, opened 2026-03-09
feat!: drop IE 11 and modernize browser targets

kdaviduik:
Part of https://github.com/Shopify/developer-tools-team/issues/1195

---

BREAKING CHANGE: Drop IE 11, Safari 8, iOS 8, Android 4.4 browser support.
Minimum browser targets are now rolling modern versions aligned with
Shopify Online Store themes. The library now emits ES2015+ JavaScript
(native classes, arrow functions, template literals) and no longer
transpiles to ES5.

- Update browserslist to union of @shopify/browserslist-config + Theme Store
  requirements
- Remove 15 Babel ES5 transform plugins + dead @babel/plugin-external-helpers
- Remove 7 polyfill imports from buybutton.js (whatwg-fetch, core-js/*)
- Replace UglifyJS with terser for ES2015+-compatible minification
- Add terser error guard against silent minification failure
- Keep @babel/plugin-transform-modules-commonjs for Browserify test pipeline
  (until PR 7a)
- Native ES2015 classes replace loose-mode Babel classes (non-enumerable
  prototype methods)
- Create major changeset for 4.0.0 version bump
- Pin pnpm/action-setup to commit SHA in ci.yml (supply chain hardening)
- Add @changesets/changelog-github as explicit devDependency (was phantom)

PR #945, opened 2026-03-09
chore: enable automatic CHANGELOG generation via Changesets

kdaviduik:
Part of https://github.com/Shopify/developer-tools-team/issues/1195

---

Switch from manual CHANGELOG management to @changesets/changelog-github.
- Update .changeset/config.json to use @changesets/changelog-github
- Pass GITHUB_TOKEN to changeset version step in release workflow
  (@changesets/changelog-github needs it to fetch PR metadata for links)
- Fix git add to include CHANGELOG.md (pre-existing bug — changeset version
  modifies CHANGELOG.md but it was never staged in the commit step)
- Fix stale v3.0.6 '(Unreleased)' label — published to npm August 25, 2025

## Verification

Smoke-tested the CHANGELOG automation locally:

1. Created a temporary test changeset (`@shopify/buy-button-js: patch`)
2. Ran `GITHUB_TOKEN=$(gh auth token) pnpm exec changeset version`
3. Confirmed CHANGELOG.md was updated with a new `## 3.0.7` section containing the test entry
4. Reverted all changes (`git checkout -- package.json CHANGELOG.md && rm .changeset/test-changelog-automation.md`)

Result: `changeset version` exits 0, plugin loads correctly, and CHANGELOG.md is generated with proper formatting.

PR #944, opened 2026-03-09
fix: run CI on all pull requests, not just those targeting main

kdaviduik:
The pull_request trigger had `branches: [main]`, which meant CI only
ran on PRs whose base branch is main. Stacked PRs (which target
intermediate branches) never got CI coverage. Remove the branch
filter so CI runs on every PR regardless of base branch.

The push trigger retains `branches: [main]` — we only want push
CI on main merges, not on every branch push.

PR #950, opened 2026-03-12
feat!: migrate ESLint 3 to ESLint 9 flat config with typescript-eslint

kdaviduik:
Part of https://github.com/Shopify/developer-tools-team/issues/1195

---

Replaces ESLint 3.3.1 + unmaintained eslint-plugin-shopify v13 with
ESLint 9 flat config + typescript-eslint v8. This enables type-aware
linting for .ts files during the ongoing TypeScript migration.

Key decisions:
- 22 TypeScript rules configured (12 PRD-specified + 10 additional
  safety rules including switch-exhaustiveness-check,
  prefer-nullish-coalescing, consistent-type-imports)
- src/types/ override suppresses known any/Function debt (deferred to PR 20)
- Test files excluded (test linting added in Phase 5)
- src/utils/ and src/styles/ now linted (18 files previously excluded
  by .eslintignore workaround)
- Removed redundant lint from test script (CI already runs lint separately)
- Used fileURLToPath workaround for __dirname (Node 18 compat)

Also fixes lint violations found by expanded coverage:
- hasOwnProperty -> Object.hasOwn() in product-set.js
- Removed unused isObject function in merge.js
- Removed unused element param in frame-utils.js
- Removed unused err binding in ui.js
- Cleaned up stale eslint-disable directives across 4 files

BREAKING CHANGE: ESLint config format changed from .eslintrc to
eslint.config.mjs. Any custom ESLint integrations must update.


kdaviduik:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/950?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#950** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/950?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/950?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#946** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/946?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#954** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#945** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/945?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#951](https://github.com/Shopify/buy-button-js/pull/951) <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/951?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#942** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#943** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#944** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#941** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#947** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

package.json
```diff
@@ -64,16 +64,18 @@
...
+    "@eslint/js": "^9.39.4",
     "@shopify/js-uploader": "https://github.com/Shopify/js-uploader.git",
     "@types/jest": "^30.0.0",
     "@types/node": "^24.5.2",
     "aws-sdk": "2.6.8",
     "babelify": "10.0.0",
     "chai": "4.2.0",
-    "eslint": "3.3.1",
-    "eslint-plugin-shopify": "13.0",
+    "core-js": "3.1.4",
```
kdaviduik: **blocking: `core-js: 3.1.4` re-added despite PR 946 removing it.**

Evidence chain:

1. `git show 381c4cb -- package.json` confirms this line was `+` added by this PR's primary commit.
2. `pnpm-lock.yaml:5422` shows `core-js@3.1.4: {}` — the empty braces mean **zero** transitive consumers. It's only present because it's declared here at the top level.
3. Grep `require\(['"]core-js|from ['"]core-js|import.*core-js` across `src/` and `test/`: **zero hits**.
4. PR 946 (commit `9e852f8`) deliberately removed `core-js: "3.1.4"` from runtime deps as part of the IE 11 drop — dropping IE 11 eliminated all polyfill needs.
5. The migration doc in this very PR still says `[x] 4.5. Remove polyfill dependencies from package.json: core-js, whatwg-fetch` (`docs/tasks/tasks-prd-typescript-migration.md:229`).
6. `pnpm-lock.yaml:1345` shows npm itself flags `core-js@<3.23.3` as deprecated: *"core-js@<3.23.3 is no longer maintained and not recommended for usage"*. Pinning to `3.1.4` adds a deprecation warning on every `pnpm install`.

Almost certainly a rebase-resolution artifact (take-theirs on a conflict where the other side was the PR 946 deletion). Suggested fix: remove from `package.json` and regenerate `pnpm-lock.yaml`.

```suggestion
```

package.json
```diff
@@ -27,9 +27,9 @@
   "scripts": {
     "start": "rm -rf tmp && mkdir tmp && pnpm run src:watch & pnpm run styles:watch & pnpm run serve",
     "build": "pnpm run clean && pnpm run styles && pnpm run images:copy && pnpm run src:transpile && pnpm run src:build",
-    "test": "pnpm run lint && pnpm run testem",
+    "test": "pnpm run testem",
```
kdaviduik: non-blocking: Lint removed from `pnpm test` creates a local-workflow regression — `pnpm test` passing locally no longer implies CI will pass.

The commit message acknowledges: *"Local workflow change: pnpm test no longer runs lint (CI runs it separately)."* CI does run lint (`ci.yml:33`), so the change is correct for CI. But for local dev, two alternatives worth considering:

- **Keep `test` as a meta-command**: `"test": "pnpm run lint && pnpm run testem"`. The "redundant with CI" argument is weak — local feedback is faster than CI.
- **Add a pre-commit hook** (husky + lint-staged) running `eslint --fix` on staged files. Scales better since it only lints changed files.

Not blocking; just a workflow-design choice worth naming.

package.json
```diff
@@ -27,9 +27,9 @@
...
     "serve": "http-server",
-    "lint": "eslint --max-warnings=0 -c .eslintrc src/*",
+    "lint": "eslint --max-warnings=0 src/",
```
kdaviduik: praise: Three good things in one line, worth unpacking:

1. **`-c .eslintrc` removed** — correct for ESLint 9 flat config, which auto-discovers `eslint.config.mjs`.
2. **`src/*` → `src/`** — cleaner and robust to filename surprises. Note: this is a simplification, *not* a coverage fix (ESLint recursed into nested directories under `src/*` already — proof: stale `no-param-reassign` disables existed in `src/components/product-set.js:26,30` before this PR).
3. **The real 18-file coverage expansion** comes from deleting `.eslintignore` (which contained `src/utils/*`, `src/styles/*`, `src/polyfills/*`). That's an **information-hiding win** worth praising separately: the old setup had lint policy split across two files with no documented rationale for the exclusions. The new setup consolidates policy into `eslint.config.mjs` — `ignores` live alongside `files` and `rules`, with inline comments where judgment is involved (the `src/types/` `TECH DEBT` note). Orphaned knowledge became colocated knowledge.

One follow-up question on the `.eslintignore` deletion → see the inline on `eslint.config.mjs:12` about `src/styles/*.js` generated files.

package.json
```diff
@@ -64,16 +64,18 @@
     "@babel/preset-typescript": "^7.27.1",
     "@changesets/changelog-github": "^0.6.0",
     "@changesets/cli": "^2.28.1",
+    "@eslint/js": "^9.39.4",
```
kdaviduik: non-blocking: With `eslint-plugin-shopify` gone, there's no replacement for its style rules. The new config uses `eslint:recommended` + type-aware rules only — correctness focus is appropriate, but **stylistic uniformity is now a gap**.

Three options worth naming explicitly:
- Accept the gap and rely on team consistency / code review.
- Add Prettier for formatting (most common modern choice).
- Add `@stylistic/eslint-plugin` (maintained replacement for stylistic-only rules).

Worth acknowledging the decision in the commit message rather than leaving it implicit.

eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+import globals from 'globals';
+import { fileURLToPath } from 'node:url';
+import { dirname } from 'node:path';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+export default tseslint.config(
+  {
+    ignores: [
+      'dist/',
```
kdaviduik: question: Are generated `src/styles/*.js` files intentionally in lint scope now?

Three files under `src/styles/` are `.js` exports of CSS strings, regenerated by `script/host-styles.js` and `script/embed-styles.js` on every build:
- `src/styles/host/host.js` (long CSS bundle)
- `src/styles/host/conditional.js`
- `src/styles/embeds/conditional.js`

The old `.eslintignore` excluded `src/styles/*` specifically because these are build outputs. Today's `eslint:recommended` rules don't trip on `export default "..."` so they lint clean by accident. But if the CSS generator ever emits output that tickles a rule (e.g., `no-useless-escape` on a regex-laden CSS string), CI would fail for a reason the author can't fix by editing the `.js` files — they'd have to fix the generator.

Options:
- **Leave as-is** (current): accept the risk.
- **Exclude under `ignores`**: add `'src/styles/**/*.js'` to this array. Explicit, preserves the original policy reason.
- **Rewrite the generator** to emit typed exports.

Non-blocking but worth naming — the `.eslintignore` exclusion existed for a reason.

eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+      'build/',
+      'test/',
+      'node_modules/',
+      '**/*.d.ts',
+    ],
+  },
+
+  {
+    files: ['src/**/*.js'],
+    ...eslint.configs.recommended,
```
kdaviduik: non-blocking: The JS-block spread pattern works today but is fragile. Two issues:

1. If `@eslint/js` ever adds a `files` property to `recommended`, the `files: ['src/**/*.js']` scoping above would be silently overridden by the spread.
2. The outer `rules` key overrides `rules` from the spread, which is why line 31 has to re-spread `eslint.configs.recommended.rules`. Known flat-config footgun.

The TS block at line 36 already uses the cleaner `extends:` pattern. Applying that same shape to the JS block would look like:

```suggestion
  {
    files: ['src/**/*.js'],
    extends: [eslint.configs.recommended],
    languageOptions: {
      globals: {
        ...globals.browser,
      },
    },
    rules: {
      'no-unused-vars': ['error', { caughtErrorsIgnorePattern: '^_' }],
    },
  },
```

This is the pattern `tseslint.config()` is designed for and matches the TS block's style. Non-blocking robustness improvement for a foundational file future migrations will build on.

eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+  },
+
+  {
+    files: ['src/**/*.ts'],
+    extends: [
+      ...tseslint.configs.recommendedTypeChecked,
+    ],
+    languageOptions: {
+      parserOptions: {
+        projectService: true,
```
kdaviduik: praise: `projectService: true` + `tsconfigRootDir: __dirname` (via `fileURLToPath(import.meta.url)`) is the modern, maintenance-light approach. Auto-discovers the TS project and handles multi-project setups without manual path maintenance. Correct for ESLint 9 + typescript-eslint v8.

Also worth a broader structural praise for the whole file: 86 lines organized into four discrete config objects (ignores, JS files, TS files, `src/types/` override). Each object has a single concern; shared knowledge (TS parser options) lives at one layer. The alternative — a single merged object with `files: ['**/*.{js,ts}']` and conditional rule suppression — would leak JS-vs-TS knowledge into every rule activation and either duplicate the `src/types/` override or require runtime branching. The current shape hides that complexity behind a flat exported array. Textbook deep module.

eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+      '@typescript-eslint/no-unsafe-assignment': 'error',
+      '@typescript-eslint/no-unsafe-member-access': 'error',
+      '@typescript-eslint/no-unsafe-call': 'error',
+      '@typescript-eslint/no-unsafe-return': 'error',
+      '@typescript-eslint/no-unsafe-argument': 'error',
+      '@typescript-eslint/no-floating-promises': 'error',
+      '@typescript-eslint/no-misused-promises': 'error',
+      '@typescript-eslint/await-thenable': 'error',
+      '@typescript-eslint/no-unnecessary-type-assertion': 'error',
+      '@typescript-eslint/consistent-type-assertions': ['error', { assertionStyle: 'never' }],
```
kdaviduik: non-blocking (forward-looking): `consistent-type-assertions: { assertionStyle: 'never' }` forbids all `as T` casts.

**Scoped to `src/**/*.ts` only**, so zero impact today (src/ has almost no `.ts` files yet, and `src/types/` has its own override). Impact starts in the TS migration (PRs 9-21).

The design question is worth stating: does the team want narrowing to be **silent** (`as T` anywhere) or **deliberate** (type predicates, `satisfies`, explicit typing)? Both are defensible.

- `'never'` is the "deliberate" choice — forces every narrowing site to be named via `function isFoo(x): x is Foo` or similar. Cost: every `JSON.parse(str)` and `unknown`-disambiguation site needs a predicate. Benefit: no silent casts hiding in the codebase.
- `['error', { assertionStyle: 'as' }]` is the pragmatic middle ground — requires `as` over angle brackets but still allows assertions. Less strict, keeps syntax consistent.

Worth confirming this is the intended policy (vs. copied from an example).

eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+
+      '@typescript-eslint/switch-exhaustiveness-check': ['error', { requireDefaultForNonUnion: true }],
+      '@typescript-eslint/prefer-nullish-coalescing': 'error',
+      '@typescript-eslint/consistent-type-imports': 'error',
+      '@typescript-eslint/no-unnecessary-condition': 'error',
+      '@typescript-eslint/return-await': ['error', 'in-try-catch'],
+      '@typescript-eslint/require-await': 'error',
+
+      '@typescript-eslint/explicit-function-return-type': 'warn',
+      '@typescript-eslint/strict-boolean-expressions': 'warn',
```
kdaviduik: non-blocking: `explicit-function-return-type` and `strict-boolean-expressions` are `warn` while the other 18 typescript-eslint rules (lines 48-66) are `error`. With `--max-warnings=0` in the lint script, warnings fail CI today — practical effect is identical. But the semantic inconsistency will confuse future readers.

The intent exists: `docs/tasks/prd-typescript-migration.md:97` says warn-level rules *"upgrade to error after Phase 4"* — so these two are transitional. But that policy lives in a separate doc, not in `eslint.config.mjs`. A brief inline comment would keep intent co-located with the rule:

```suggestion
      // Transitional: 'warn' until Phase 4; upgrade to 'error' after all .ts files converted.
      '@typescript-eslint/explicit-function-return-type': 'warn',
      '@typescript-eslint/strict-boolean-expressions': 'warn',
```

Small information-hiding improvement — the rule and its rationale become one conceptual unit.

PR #951, opened 2026-03-14
feat: switch to changesets release PR workflow

kdaviduik:
Part of https://github.com/Shopify/developer-tools-team/issues/1195

---

## Summary

- Delete `DEPLOYING.md` (info is now in CONTRIBUTING.md instead))
  - Almost all of the previous information in DEPLOYING.md was no longer accurate now that there's a changeset release PR rather than a release happening upon merging into main
- Create `CLAUDE.md` and `AGENTS.md` as symlinks to `CONTRIBUTING.md` so that we have a single source of truth for all of this information and agents automatically "see" it
- Rewrite `npm-release.yml` for the changesets release PR pattern: pending changesets create a `[ci] release` PR; merging that PR triggers npm publish

## Why

The previous workflow published to npm immediately on PR merge. This made it extremely un-ergonomic to batch multiple merchant-facing changes in a single release, and required workarounds (like adding `[DOCS]` to the PR title) to avoid releasing a new Buy Button JS version when there were zero code changes. The release PR pattern (used by Hydrogen and many Shopify repos) creates an intermediate PR showing exactly what will be published and is the more modern, preferred way of managing releases.

## External prerequisite

`SHOPIFY_GH_ACCESS_TOKEN` has been provisioned via [github-actions-access-provider#2613](https://github.com/Shopify/github-actions-access-provider/issues/2613)

PR #943, opened 2026-03-09
fix: remove Firefox requirement from testem CI config and fix view tests

kdaviduik:
Testem's launch_in_ci required both Chrome and Firefox, but Firefox
is not installed locally, which caused `testem ci` to abort before
running any tests. Remove Firefox — Chrome-only is sufficient for
local test validation. The entire test runner (Mocha + Testem) is
being replaced by Vitest + happy-dom in a future PR.

Also fix 6 pre-existing test failures in test/unit/view.js where
tests tried to reassign the read-only ES module import `Iframe`
to spy on the constructor. Replace with assertions on the
observable outcome (view.iframe.el.getAttribute('title')), which
is both correct under modern strict mode and a stronger test
pattern — verifying behavior rather than implementation details.

PR #942, opened 2026-03-09
feat: migrate Yarn v1 to pnpm

kdaviduik:
Replace Yarn v1 with pnpm 10 as the package manager.

- Create .npmrc (enable-pre-post-scripts, shamefully-hoist)
- Generate pnpm-lock.yaml via pnpm import (exact version preservation)
- Update all package.json scripts from yarn/npm to pnpm
- Rename prepublish to prepublishOnly (modern lifecycle hook)
- Add packageManager field for Corepack enforcement
- Update CI workflows with pnpm/action-setup and --frozen-lockfile
- Update dev.yml, shipit.production.yml, readme.md, DEPLOYING.md
- Fix changeset detection in npm-release.yml (check *.md not dir empty)
- Delete yarn.lock

shamefully-hoist is temporary: needed until browserify is replaced
by Vitest (PR 7a). enable-pre-post-scripts is permanent: the project
uses pretest and prepublishOnly lifecycle hooks.

PR #941, opened 2026-03-09
docs: revise TypeScript migration PRD and task list for tech stack modernization

kdaviduik:
## Summary

- Rewrites the TypeScript migration PRD and task list to expand scope from TS-only migration to full tech stack modernization
- Adds Phase 3: Tooling Modernization (PRs 3-8) — pnpm, browser target drop, ESLint 9, Vite, Vitest, dependency cleanup — before TS file conversion begins
- Fixes critical issues: wrong test framework types (`@types/jest` → Vitest), 14 missing files, dependency-tier ordering violations, npm/yarn inconsistency
- Adds complete 49-file source inventory with tier classification and 30-file test inventory with PR assignments
- Documents pre-existing bugs, CSS class type mismatches, and `Record<string, any>` debt for systematic cleanup

## Key decisions

- **Major version bump 3.0.6 → 4.0.0** for IE 11 drop (breaking change for legacy browser consumers)
- **Tooling modernization BEFORE file conversion** — modern tools make migration faster, safer, and better enforced
- **~26 PRs** across 3 phases: tooling (7 PRs), source conversion (13 PRs), test conversion (6 PRs)
- **Modal updater moved to PR 17** — cross-tier dependency on Product component

PR #952, opened 2026-03-14
feat: migrate npm-release.yml to OIDC trusted publishing

kdaviduik:
Part of https://github.com/Shopify/developer-tools-team/issues/1195

---

## Summary

- Migrate npm-release.yml from static `NPM_TOKEN` to OIDC trusted publishing
- Fix `package.json` repository field from SSH shorthand to structured HTTPS format (required for provenance attestation, which is a prerequisite for OIDC trusted publishing)

## Why

OIDC replaces long-lived `NPM_TOKEN` secrets with short-lived tokens derived from the GitHub Actions workflow identity. This eliminates the risk of credential compromise — tokens are scoped to the specific workflow run and cannot be reused. Standard at Shopify (Hydrogen, CLI, theme-tools, flash-list).

npm classic tokens were revoked December 9, 2025. This repo's `NPM_TOKEN` was a classic token, so it's already dead and this migration is a necessity in order to do future npm releases.

## How OIDC works

1. `id-token: write` permission allows the workflow to request an OIDC token from GitHub
2. `setup-node` with `registry-url` configures `.npmrc` for auth
3. `NODE_AUTH_TOKEN` is NOT set — this causes npm to fall through to OIDC token exchange
4. `NPM_TOKEN: ''` (empty string) explicitly forces OIDC fallback
5. npm 11 (bundled with Node 24) natively supports OIDC token exchange with npmjs.com

## Snapit

(similar to Hydrogen) Snapit OIDC migration is **deferred**. Only `npm-release.yml` is configured as a Trusted Publisher on npmjs.com. To get `/snapit` to work we would need to move the contents of the snapit.yml into npm-release.yml.
- We do not need `/snapit` to test our changes and verify everything works. We can instead just manually build this package and use that rather than the npm release from `/snapit`

### DIFF
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -20,7 +20,7 @@ jobs:
     permissions:
       contents: write      # push version commits, create GitHub releases
       pull-requests: write # create/update the "[ci] release" PR
-      id-token: write      # OIDC for npm provenance attestation
+      id-token: write      # OIDC for npm trusted publishing AND provenance attestation
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: "18.20"
+          node-version: "24"
           cache: 'pnpm'
           registry-url: 'https://registry.npmjs.org'
 
@@ -56,6 +56,8 @@ jobs:
           # scoped to this repo only (contents:write, pull-requests:write),
           # rotated every 2 hours via github-actions-access-provider.
           GITHUB_TOKEN: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          # Empty NPM_TOKEN forces npm to fall through to OIDC token exchange.
+          # Do NOT set NODE_AUTH_TOKEN — setup-node's registry-url writes it
+          # into .npmrc; leaving it undefined triggers the OIDC flow.
+          NPM_TOKEN: ''
           NPM_CONFIG_PROVENANCE: true
--- a/package.json
+++ b/package.json
@@ -12,7 +12,10 @@
   ],
   "description": "BuyButton.js allows merchants to build Shopify interfaces into any website",
   "main": "lib/buybutton.umd.js",
-  "repository": "git@github.com:Shopify/buy-button-js.git",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/buy-button-js.git"
+  },
   "publishConfig": {
     "access": "public",
     "@shopify:registry": "https://registry.npmjs.org/",


kdaviduik:
* **#952** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#951** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#945** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/945?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#954](https://github.com/Shopify/buy-button-js/pull/954) <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/954?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#942** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#943** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#944** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#941** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#947** <a href="https://app.graphite.com/github/pr/Shopify/buy-button-js/947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.github/workflows/npm-release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     permissions:
       contents: write      # push version commits, create GitHub releases
       pull-requests: write # create/update the "[ci] release" PR
-      id-token: write      # OIDC for npm provenance attestation
+      id-token: write      # OIDC for npm trusted publishing AND provenance attestation
```
kdaviduik: praise: Good update to this comment. `id-token: write` now serves two purposes (OIDC trusted publishing AND provenance attestation); both use the same token-exchange primitive, so the permission scope is shared. Future maintainers who see this line won't wonder why it's still needed after the `NPM_TOKEN` removal.

.github/workflows/npm-release.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: "18.20"
+          node-version: "24"
```
kdaviduik: praise: Node 24 is necessary here — **npm 11.5+** (bundled with Node 24) is the first npm version with native OIDC token exchange for npmjs.com trusted publishing. Matches Hydrogen's rationale comment: `node-version: 24 # Needed for npm@11 for Trusted Publishing` ([permalink](https://github.com/Shopify/hydrogen/blob/2fa1108bf899377a949f612694fd8c67b5151a77/.github/workflows/release.yml#L58)).

nit: Could be worth adding a similar inline comment here (`# Required: npm 11.5+ bundled with Node 24 enables OIDC trusted publishing`) so the version choice is self-documenting — but not blocking.

.github/workflows/npm-release.yml
```diff
@@ -56,6 +56,8 @@ jobs:
           # scoped to this repo only (contents:write, pull-requests:write),
           # rotated every 2 hours via github-actions-access-provider.
           GITHUB_TOKEN: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NPM_CONFIG_PROVENANCE: true
\ No newline at end of file
+          # Empty NPM_TOKEN forces npm to fall through to OIDC token exchange.
+          # Do NOT set NODE_AUTH_TOKEN — setup-node's registry-url writes it
+          # into .npmrc; leaving it undefined triggers the OIDC flow.
```
kdaviduik: praise: This inline comment is exactly the kind of tribal-knowledge-becomes-documentation that prevents a future maintainer from "cleaning up" what looks like a mistake. The `NPM_TOKEN=''` idiom is counterintuitive enough that a drive-by refactor would be tempting without this note.

non-blocking: The PR's inline comment says "leaving it undefined triggers the OIDC flow" — directionally right, but the actual mechanism is more nuanced. `actions/setup-node` with `registry-url` writes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}` into `.npmrc` ([authutil.ts L43-L46 at the pinned SHA](https://github.com/actions/setup-node/blob/49933ea5288caeca8642d1e84afbd3f7d6820020/src/authutil.ts#L43-L46)), and when `NODE_AUTH_TOKEN` is unset, setup-node exports the literal placeholder `XXXXX-XXXXX-XXXXX-XXXXX` ([authutil.ts L49-L53](https://github.com/actions/setup-node/blob/49933ea5288caeca8642d1e84afbd3f7d6820020/src/authutil.ts#L49-L53)). So `.npmrc` ends up with `_authToken=XXXXX-XXXXX-XXXXX-XXXXX`; npm classic-token auth attempts it, fails, and npm 11.5+ falls through to OIDC trusted publishing. Not "no auth token" — rather "bogus auth token that triggers OIDC fallback." Worth tightening the comment if you feel like it, since the precise mechanism is what a future debugger will need.

.github/workflows/npm-release.yml
```diff
@@ -56,6 +56,8 @@ jobs:
...
+          NPM_TOKEN: ''
```
kdaviduik: praise: `NPM_TOKEN: ''` matches Hydrogen's production pattern verbatim ([permalink](https://github.com/Shopify/hydrogen/blob/2fa1108bf899377a949f612694fd8c67b5151a77/.github/workflows/release.yml#L141)), where the same combination successfully publishes with OIDC across three flows (latest, next, back-fix). The empty string is load-bearing — it ensures `changesets/action` doesn't interpret `NPM_TOKEN` as set and write a real auth token into `.npmrc`.

non-blocking: **Consider adding a Test OIDC Token pre-flight step** before the `Create Release Pull Request or Publish` step. Hydrogen does this on all three publish flows ([L122-L127](https://github.com/Shopify/hydrogen/blob/2fa1108bf899377a949f612694fd8c67b5151a77/.github/workflows/release.yml#L122-L127) unconditional on back-fix at [L269](https://github.com/Shopify/hydrogen/blob/2fa1108bf899377a949f612694fd8c67b5151a77/.github/workflows/release.yml#L269), conditional on latest / next):

```yaml
- name: Test OIDC Token
  run: |
    TOKEN=$(curl -s -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
      "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=npm" | jq -r '.value')
    [ -n "$TOKEN" ] && echo "✅ OIDC working" || exit 1
```

The `$ACTIONS_ID_TOKEN_REQUEST_TOKEN` and `$ACTIONS_ID_TOKEN_REQUEST_URL` env vars are auto-injected by the Actions runtime when `id-token: write` is set. Benefit: if the npmjs.com trusted-publisher config is misconfigured (wrong workflow path, wrong branch, wrong org), you get a clean 1-line "❌ OIDC not working" signal instead of a cryptic 401 from npm mid-publish. Costs ~1s per run.

.github/workflows/npm-release.yml
```diff
@@ -56,6 +56,8 @@ jobs:
...
+          NPM_CONFIG_PROVENANCE: true
```
kdaviduik: nit: Both `NPM_CONFIG_PROVENANCE: true` (env var here) and `publishConfig.provenance: true` (in `package.json` line 22) are set. Harmless duplication — either one forces provenance on publish — but a future reader may wonder why both are needed. Not worth changing; just flagging in case you want to pick one source of truth.

package.json
```diff
@@ -12,7 +12,10 @@
   ],
   "description": "BuyButton.js allows merchants to build Shopify interfaces into any website",
   "main": "lib/buybutton.umd.js",
-  "repository": "git@github.com:Shopify/buy-button-js.git",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/buy-button-js.git"
```
kdaviduik: praise: Required change for provenance attestation. npm's provenance validator parses `repository.url` to construct the source repo slug recorded in the attestation; the previous SSH shorthand (`git@github.com:Shopify/buy-button-js.git`) doesn't parse as a valid git URL per npm's parser, but the `git+https://…/…git` form does.

Confirmed this matches Hydrogen's `repository` field. (flash-list uses a plain `https://github.com/Shopify/flash-list` form — also valid, npm accepts both — but Hydrogen's `git+https://` form is the most explicit and what npm's own docs use in examples.)

EvilGenius13 (approved):

PR #955, opened 2026-08-22
chore: bump node to latest LTS + GitHub actions to latest

kdaviduik:
## Context

Node 18 reached end of life in April 2025. This bumps all node version references in the repo to v24.19.0 (Krypton), the current active LTS.

Pinned all GitHub Actions to their latest versions by commit SHA (supply-chain best practice, matching Hydrogen's pattern):
  - actions/checkout → v7.0.1
  - actions/setup-node → v7.0.0
  - pnpm/action-setup → v6.0.10

Also adds a new CI job (same as Hydrogen's) to ensure that the node versions in `dev.yml` and `.nvmrc` can't go out of sync without CI screaming.

### Verifying the pinned SHAs

Each SHA below links to its GitHub commit page so you can confirm it matches the tagged release:

| Action | Version | Commit SHA | Verify |
|--------|---------|------------|--------|
| `actions/checkout` | v7.0.1 | `3d3c42e5aac5ba805825da76410c181273ba90b1` | [commit](https://github.com/actions/checkout/commit/3d3c42e5aac5ba805825da76410c181273ba90b1) / [release](https://github.com/actions/checkout/releases/tag/v7.0.1) |
| `actions/setup-node` | v7.0.0 | `820762786026740c76f36085b0efc47a31fe5020` | [commit](https://github.com/actions/setup-node/commit/820762786026740c76f36085b0efc47a31fe5020) / [release](https://github.com/actions/setup-node/releases/tag/v7.0.0) |
| `pnpm/action-setup` | v6.0.10 | `0977fd99725f1db4007ccb2928dbb4e90d06cc86` | [commit](https://github.com/pnpm/action-setup/commit/0977fd99725f1db4007ccb2928dbb4e90d06cc86) / [release](https://github.com/pnpm/action-setup/releases/tag/v6.0.10) |

## Why dev.yml can't just read from .nvmrc

The Shopify dev tool requires an explicit exact version in dev.yml — the version field is required and is used to run nvm install <version>. It can't read from .nvmrc. So we can't eliminate the version specification in dev.yml and rely on .nvmrc alone. This is why the two files exist separately and why the sync-check CI job is needed: .nvmrc uses the flexible major version v24 (for nvm) while dev.yml pins the exact v24.19.0 (for the dev tool).

The dev tool does its own .nvmrc ↔ dev.yml check during dev up, but that only runs when a contributor remembers to run it locally. The new CI job catches drift on every PR — including from external contributors who don't have the dev tool.

This mirrors the pattern already in place in Shopify/hydrogen.


.github/workflows/ci.yml
```diff
@@ -6,21 +6,94 @@ on:
   pull_request:
 
 jobs:
+  check-node-version:
```
EvilGenius13, line 9: This is like the hydrogen check too right?

EvilGenius13 (approved):

.github/workflows/ci.yml
kdaviduik: yep [exactly](https://github.com/Shopify/hydrogen/blob/2a2738ba20487ccc07006815fe40e93b24cb5f08/.github/workflows/ci.yml#L9-L80)!

PR #956, opened 2026-08-22
fix: make dev up, dev server, and dev docs work

kdaviduik:
## Why

`dev up` was broken (the `node` task used a `pnpm:` key that current dev rejects), `dev server` failed to bind port 8080 (devns conflict), and `dev docs` had no Ruby/Jekyll toolchain provisioned. This PR fixes all three so the standard dev workflow works out of the box — no manual setup steps.

## What

**`dev up` — fixed node task + provisioned Jekyll**
- `dev.yml` `node` task: `pnpm: 10.16.1` → `package_manager: pnpm@10.16.1` + `packages: [.]` (the old `pnpm:` key was rejected by dev's validator)
- Added `ruby` + `bundler` to `up:` — provisions Ruby 3.3.6 and Jekyll 4.4.1 so `dev docs` works without manually installing Jekyll
- New files: `.ruby-version` (3.3.6), `Gemfile` (`gem 'jekyll', '~> 4.4'`), `Gemfile.lock` (generated by `dev up`)

**`dev server` — renamed command, fixed port binding, removed manual setup**
- Renamed `start` → `server` in `dev.yml` (matches the convention across Shopify repos)
- `server` now runs `pnpm run build && pnpm run start` — a restart picks up source changes cleanly
- Renamed `open: example` → `open: app` — Ctrl-T during `dev server` is hardcoded to look for a link named `app`
- Fixed `EADDRINUSE` on port 8080: changed `serve` script from `http-server` to `http-server -a 127.0.0.1 -p 8080`

  `devns` (dev's networking daemon) holds `0.0.0.0:8080` in a CLOSED socket state to proxy `*.shop.dev` traffic. `http-server` defaults to binding `0.0.0.0:8080`, which conflicts. Binding to `127.0.0.1` avoids the conflict. `-p 8080` is also needed — without it, `http-server` uses `portfinder` which checks `0.0.0.0` (still blocked by devns) and auto-increments to 8081+.

  Evidence: [http-server README](https://github.com/http-party/http-server#readme) documents that `-a` defaults to `0.0.0.0` and `-p` defaults to `8080`. You can reproduce the conflict by running `node -e "require('http').createServer().listen(8080, '0.0.0.0')"` — it throws `EADDRINUSE`, while `listen(8080, '127.0.0.1')` succeeds.

  **Why not `-p 0` (auto-find open port)?** The http-server docs suggest `-p 0` to let `portfinder` find an open port starting at 8080. However, `portfinder` checks availability on `0.0.0.0` (where devns holds the CLOSED socket), not on the `-a 127.0.0.1` address — so it sees 8080 as taken and skips to 8081, which breaks `dev open app` (hardcoded to `http://localhost:8080/`). Using an explicit `-p 8080` bypasses `portfinder` entirely and binds directly to `127.0.0.1:8080`.

- Renamed `index.example.html` → `index.html` and hardcoded the hydrogen-preview demo store credentials (same public credentials already used in `test-manual/index.html`). Removed `/index.html` from `.gitignore` so it's committed. This eliminates the manual `cp` + credential-filling step — `dev server` just works.

  **Why this is safe:** `index.html` was originally gitignored in 2016 because it contained a real Storefront API key. The example/template pattern was the 2016 solution to "don't commit API keys." This concern no longer applies: storefront access tokens are public-by-design (they're embedded in browser-side JS on merchant websites — that's literally what buy-button-js does), they only grant read access to products/collections, and the specific credentials used here are already committed in `test-manual/index.html` in this same repo.

**`dev docs` — fixed docs command + stale docs**
- `docs` script: `jekyll serve` → `bundle exec jekyll serve` (uses the Jekyll provisioned by `dev up`)
- `docs/readme.md`: was copy-pasted from the `js-buy-sdk` repo (wrong title, wrong URL, referenced non-existent `doc-build`/`doc-serve` scripts) — rewritten to accurately describe this repo's docs setup
- `readme.md`: Documentation section no longer says to manually `gem install jekyll` (dev up handles it); corrected "gh-pages from main" to "Pages builds /docs from main" (no gh-pages branch); removed manual `cp index.example.html index.html` step

**Ignore files**
- `.gitignore`: removed `/index.html` (now committed); added `.jekyll-cache` (Jekyll 4 cache dir)
- `.npmignore`: excluded `/Gemfile`, `/Gemfile.lock`, `/.ruby-version` so Ruby tooling doesn't ship to npm consumers (`index.html` was already excluded)

## How to verify

```bash
dev up          # should converge cleanly (Ruby + Jekyll installed)
dev server      # should build, then serve at http://localhost:8080/ (demo page with hydrogen-preview store)
dev docs        # should serve docs at http://localhost:4000/buy-button-js/
```


Gemfile.lock
```diff
@@ -0,0 +1,106 @@
+GEM
+  remote: https://rubygems.org/
```
stephanie-shopify, line 2: should this file behere?

stephanie-shopify (approved):

index.html
```diff
@@ -9,14 +9,19 @@
     <script src="tmp/buybutton.dev.js"></script>
     <script data-shopify-buy-ui>
       var client = ShopifyBuy.buildClient({
-        domain: '', // ex 'storename.myshopify.com'
-        storefrontAccessToken: '' // previously 'apiKey', now deprecated
+        domain: 'hydrogen-preview.myshopify.com',
+        storefrontAccessToken: '79921ccabf3be126fbd83c8fa0e04a3c'
```
EvilGenius13, line 13: I assume it's fine having this as a public thing?

EvilGenius13 (approved):

EvilGenius13:
Forgot to say I 🎩 and everything booted up and worked.

index.html
kdaviduik: yep! I had a comment about this in the PR description 
image

Gemfile.lock
kdaviduik: Yes - jekyll only exists as a Ruby gem. Before, the instructions were to manually install it. But it's a smoother developer experience to just have it installed automatically as part of `dev up`, so this PR wires that up, which also requires adding the ruby version file and the Gemfile.lock since those didn't exist before. 

image

PR #957, opened 2026-09-11
Publish to npm with OIDC trusted publishing

graygilmore:
Both publishing workflows in this repo still authenticate to npm with the `NPM_TOKEN` secret. npm's recommended path for CI publishing is now [trusted publishing](https://docs.npmjs.com/trusted-publishers) (OIDC), which is what hydrogen and theme-tools already use, so this PR moves both workflows over:

- `npm-release.yml`: sets `NPM_TOKEN: ''` so `changesets/action` skips writing a token to `.npmrc` and npm falls through to OIDC. `changesets/action` moves to v1.9.0 because that path only exists from v1.7.0. `setup-node` gets `registry-url` so the CLI knows which registry to exchange the OIDC token with.
- `snapit.yml`: bumps `Shopify/snapit` to v0.1.0, which dropped `NPM_TOKEN` support and publishes over OIDC. The `github_comment_included_packages` input was renamed to `comment_packages` in that release.

## Before merging

Trusted publishing is configured per package *and* per workflow file on npmjs.com, so `@shopify/buy-button-js` needs both `npm-release.yml` and `snapit.yml` registered for this repo before this lands. Otherwise the next publish fails with `ENEEDAUTH`. That happens on the npm side, not in this diff.

Once the first OIDC publish goes through, the `NPM_TOKEN` repo secret can be deleted.

## Reviewer notes

This overlaps with #952, which makes the same `npm-release.yml` change as part of a larger modernization stack. This PR is intentionally standalone against `main` so it can land on its own. Happy to drop it if that stack is close to landing.

`issue_comment` workflows always run from `main`, so the `snapit.yml` change can't be tried on this PR. The `Force snapshot` step also still writes to `snapshot.txt` instead of `.changeset/`, which I don't think does anything; left alone since it's unrelated.


.github/workflows/npm-release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     permissions:
       contents: write # to be able to publish a GitHub release
       pull-requests: write # to be able to comment on released pull requests
-      id-token: write # to enable use of OIDC for npm provenance
+      id-token: write # to enable OIDC trusted publishing to npm
```
kdaviduik, line 21: I have a [draft PR open for this](https://github.com/Shopify/buy-button-js/pull/952/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R15-R18) already, and this PR is also missing [this change](https://github.com/Shopify/buy-button-js/pull/952/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R15-R18) in the package.json which is required for OIDC trusted publishing

kdaviduik: actually I think the change I mentioned might be required for providence not _specifically_ OIDC trusted publishing but publishing with providence is a prerequisite for OIDC trusted publishing so regardless the end result is the same - we need that code for this to be able to publish

fredericoo: cool good to know! can you take over with #952 instead and we close this one?

kdaviduik:
https://github.com/Shopify/buy-button-js/pull/952 merged

.github/workflows/npm-release.yml
kdaviduik: done!

